PR #29, opened 2023-03-02
fix: malformed requests could cause the server to panic

bradegler:


mattrandallbecker (approved):

pkg/server/token_minter.go
```diff
@@ -163,12 +163,21 @@ func (s *TokenMintServer) writeAuditLog(ctx context.Context, auditEvent *auditEv
 		return fmt.Errorf("error marshaling config: %w", err)
 	}
 
+	resourceName := "not specified"
+	if auditEvent.Request != nil {
+		resourceName = strings.Join(auditEvent.Request.Repositories, ",")
```
verbanicm, line 168: Isnt this required? Should we just bail with "Repositories is a required field"

verbanicm: nvm

verbanicm (approved):

pkg/server/token_minter.go
bradegler: If the request is poorly formed it might not have Repositories but we still want to write an audit log.

PR #32, opened 2023-03-02
feat: validation for the requested repositories - closes #30

### LINKED ISSUE #30: Support `*` for repositories input
### TL;DR

Allow a token to request all repositories with the `*` syntax.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

bradegler:
Added ability to match on a * wildcard or via a prefix match

Fixed a few typos in the configuration files

### DIFF
--- a/configs/abcxyz/github-token-minter.yaml
+++ b/configs/abcxyz/github-token-minter.yaml
@@ -1,6 +1,6 @@
 # abcxyz/github-token-minter.yaml
 - 
   if: 'assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml")'
-  respositories: ['github-token-minter']
+  repositories: ['github-token-minter']
   permissions:
     issues: 'read'
--- a/configs/abcxyz/github-token-minter.yaml
+++ b/configs/abcxyz/github-token-minter.yaml
@@ -1,6 +1,7 @@
 # abcxyz/github-token-minter.yaml
 - 
   if: 'assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml")'
-  repositories: ['github-token-minter']
+  repositories: 
+    - 'github-token-minter'
   permissions:
     issues: 'read'
--- a/configs/abcxyz/infra.yaml
+++ b/configs/abcxyz/infra.yaml
@@ -1,12 +1,12 @@
 # abcxyz/infra.yaml
 - if: 'assertion.workflow_ref.startsWith("abcxyz/infra/.github/workflows/terraform.yml")'
-  respositories:
-    - 'abcxyz/github-metrics-aggregator'
-    - 'abcxyz/github-token-minter'
-    - 'abcxyz/guardian'
-    - 'abcxyz/infra'
-    - 'abcxyz/jvs'
-    - 'abcxyz/lumberjack'
-    - 'abcxyz/pmap'
+  repositories:
+    - 'github-metrics-aggregator'
+    - 'github-token-minter'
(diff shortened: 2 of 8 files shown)


configs/abcxyz/github-token-minter.yaml
```diff
@@ -1,6 +1,6 @@
 # abcxyz/github-token-minter.yaml
 - 
   if: 'assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml")'
-  respositories: ['github-token-minter']
+  repositories: ['github-token-minter']
```
verbanicm, line 4: stick with expanded form when not empty:

```
repositories: 
  - 'github-token-minter'
```

bradegler: ack

pkg/server/token_minter.go
```diff
@@ -292,6 +299,48 @@ func (s *TokenMintServer) processRequest(r *http.Request, auditEvent *auditEvent
...
+	// Return the set of repositories that are allowed. If no repositories are requested
+	// this will result in an empty list which will work for organization level updates
+	// such as adding members to teams.
+	return repositories, nil
+}
+
+func matchesAllowed(allow string, repo string) bool {
+	switch {
+	// Matches all repositories in the allow list
+	case repo == "*":
```
verbanicm, line 331: NIT

I might change repo to request or something specific to context, this confused me to which was the star at first.

Same wiht line 309 maybe?

pkg/server/token_minter.go
```diff
@@ -292,6 +299,48 @@ func (s *TokenMintServer) processRequest(r *http.Request, auditEvent *auditEvent
...
+// validateRepositories checks the set of requested repositories against the allow list
+// to verity that it is authorizaed. Response contains the list of allowed repositories
+// after wild card match expansion.
+func validateRepositories(allowed, requested []string) ([]string, error) {
+	repositories := []string{}
+	// Loop through all of the requested repositories to verifiy that are in the configured
+	// allow list
+	for _, repo := range requested {
+		matched := false
+		for _, allow := range allowed {
```
verbanicm, line 311: Should we let allow be *?

pkg/server/token_minter.go
bradegler: done

pkg/server/token_minter.go
bradegler: Personally I feel we need to be explicit here so that new repositories aren't automatically included in the allow list.

verbanicm: sounds good.

configs/abcxyz/infra.yaml
```diff
@@ -1,12 +1,12 @@
...
-    - 'abcxyz/lumberjack'
-    - 'abcxyz/pmap'
+  repositories:
+    - 'github-metrics-aggregator'
+    - 'github-token-minter'
+    - 'guardian'
+    - 'infra'
+    - 'jvs'
+    - 'lumberjack'
+    - 'pmap'
```
sethvargo, line 10: This is an interesting change. How does org come into play now?

sethvargo (approved):

configs/abcxyz/infra.yaml
bradegler: The org is part of the installation id so you can't actually pick a repository outside of `abcxyz`. Since we only support a single install we don't need to specify.

sethvargo: Got it. I guess that makes sense.

PR #44, opened 2023-03-30
fix: updated module to handle IAM member information for the dataset correctly

bradegler:


sethvargo (approved):

terraform/variables.tf
```diff
@@ -58,18 +58,9 @@ variable "dataset_id" {
...
-    owners  = list(string)
-    editors = list(string)
-    viewers = list(string)
-  })
-  default = {
-    owners  = []
-    editors = []
-    viewers = []
-  }
+variable "dataset_editor_service_account_member" {
```
verbanicm, line 61: This is kind of specific, what if you need to use this same service account for other permissions other than dataset later?

verbanicm:
Should we also factor out service_iam from the cloud run module?

terraform/variables.tf
verbanicm: This module requires that these service accounts have these permissions right? Seems like we should be explicit about setting what token minter needs.

bradegler:
> Should we also factor out service_iam from the cloud run module?

Ugh. I guess that should be done first but that's going to be an ugly change with a lot of downstream repercussions if there are consumers that haven't pinned to a commit.

terraform/variables.tf
bradegler: Eh, I guess. I renamed it to be more generic.

verbanicm (approved):

PR #54, opened 2023-08-28
feat: add infra template for bootstrapping GTM with abc CLI

gjonathanhong:


abc.templates/infra/main.tf
```diff
@@ -0,0 +1,48 @@
...
+  source = "git::https://github.com/abcxyz/github-token-minter.git//terraform/modules/github-wif?ref=c6c2a47034268b472dbf1e51c28118c5cedb61e4"
+
+  project_id = local.project_id
+
+  id = "token-minter"
+  github = {
+    owner_id   = "REPLACE_GITHUB_OWNER_ID"
+    owner_name = "REPLACE_GITHUB_OWNER_NAME"
+    repo_id    = "REPLACE_GITHUB_REPO_ID"
+    repo_name  = "REPLACE_GITHUB_REPO_NAME"
```
gjonathanhong, line 30: @verbanicm  These repo values are only used within the module for setting `default_wif_attribute_condition`. However, because `wif_attribute_condition` is set, these values are not used, yet required. Should we make these optional?

abc.templates/infra/main.tf
```diff
@@ -0,0 +1,48 @@
...
+
+  project = local.project_id
+
+  service                    = each.value
+  disable_on_destroy         = false
+  disable_dependent_services = false
+}
+
+module "github_token_minter_github_wif" {
+  source = "git::https://github.com/abcxyz/github-token-minter.git//terraform/modules/github-wif?ref=c6c2a47034268b472dbf1e51c28118c5cedb61e4"
```
gjonathanhong, line 21: @verbanicm In the design, I wrote that modules should be colocated within the source repo of the service, but having a copy of github-wif module in each service seems difficult in terms of maintenance/consistency. Should we have separate templates for common modules that can be rendered in the abcxyz-services repo?

gjonathanhong: Combined into main github-token-minter module

terraform/main.tf
```diff
@@ -54,7 +59,12 @@ module "cloud_run" {
   ingress               = var.enable_gclb ? "internal-and-cloud-load-balancing" : "all"
   secrets               = ["github-application-id", "github-installation-id", "github-privatekey"]
   service_account_email = google_service_account.run_service_account.email
-  service_iam           = var.service_iam
+  service_iam = {
+    admins     = var.service_iam.admins
+    developers = toset(flatten([var.ci_service_account_member, var.service_iam.developers]))
```
verbanicm, line 64: I believe this is missing brackets, but also this reads funny to me, instead I would concatenate the values:

```terraform
    developers = toset(concat(var.service_iam.developers, [var.ci_service_account_member]))
```

terraform/github-wif.tf
```diff
@@ -0,0 +1,92 @@
...
+    "attribute.actor" : "assertion.actor"
+    "attribute.aud" : "assertion.aud"
+    "attribute.event_name" : "assertion.event_name"
+    "attribute.environment" : "assertion.environment"
+    "attribute.repository" : "assertion.repository"
+    "attribute.repository_id" : "assertion.repository_id"
+    "attribute.repository_owner_id" : "assertion.repository_owner_id"
+    "attribute.repository_visibility" : "assertion.repository_visibility"
+    "attribute.workflow" : "assertion.workflow"
+    "attribute.workflow_ref" : "assertion.workflow_ref"
```
verbanicm, line 29: We can remove the unnecessary values from this:
 
```suggestion
    "google.subject" : "assertion.sub"
    "attribute.actor" : "assertion.actor"
    "attribute.aud" : "assertion.aud"
    "attribute.repository_owner_id" : "assertion.repository_owner_id"
```

terraform/variables.tf
```diff
@@ -71,13 +75,45 @@ variable "dataset_id" {
...
+  type        = map(string)
+  description = "(Optional) Workload Identity Federation provider attribute mappings. Defaults to base mapping for default attribute condition."
+  default     = null
+}
+
+variable "wif_attribute_condition" {
+  type        = string
+  description = "(Optional) Workload Identity Federation provider attribute condition. Appended to base condition, matching GitHub owner and repository id. Defaults to preventing `pull_request_target` event and matching GitHub owner and repository id."
+  default     = null
+}
```
verbanicm, line 119: These can be deleted per comment above

terraform/variables.tf
```diff
@@ -71,13 +75,45 @@ variable "dataset_id" {
...
+
+variable "wif_id" {
+  description = "An ID for the Workload Identity Federation resources."
+  type        = string
+  validation {
+    condition     = length(var.wif_id) <= 22
+    error_message = "ERROR: id must be 22 characters or less."
+  }
+}
```
verbanicm, line 96: This can be deleted per comment above

terraform/github-wif.tf
```diff
@@ -0,0 +1,92 @@
...
+    issuer_uri = "https://token.actions.githubusercontent.com"
+  }
+}
+
+resource "google_service_account" "wif_service_account" {
+  project = var.project_id
+
+  account_id   = "gh-${var.wif_id}-sa" # 30 characters
+  display_name = "GitHub WIF ${var.wif_id} service account"
+}
```
verbanicm, line 85: Lets take an opinion here and create unique ids and names specific for this use case. The reason the other module is so flexible is because it serves several use cases, this wif setup is specific.

For example, the service account name should have something related to github-token-minter so we know what it is.

abc.templates/infra/outputs.tf
```diff
@@ -0,0 +1,19 @@
...
+
+output "run_service_name" {
+  description = "The Cloud Run service name."
+  value       = module.github_token_minter.run_service_name
+}
+
+output "run_service_url" {
+  description = "The Cloud Run service url."
+  value       = module.github_token_minter.run_service_url
+}
```
verbanicm, line 19: Add the workload identity values here as well

abc.templates/infra/backend.tf
```diff
@@ -0,0 +1,6 @@
+terraform {
+  backend "gcs" {
+    bucket = "REPLACE_BUCKET_NAME"
+    prefix = "REPLACE_BUCKET_PREFIX"
+  }
+}
```
verbanicm, line 6: Its best to be explicit about the providers you need, see https://github.com/abcxyz/gcp-org-terraform-template/blob/main/templates/deployment/terraform.tf#L23-L31

Same goes for terraform version

abc.templates/infra/main.tf
```diff
@@ -0,0 +1,23 @@
...
+  project_id                        = "REPLACE_PROJECT_ID"
+  automation_service_account_member = "REPLACE_AUTOMATION_SERVICE_ACCOUNT_MEMBER"
+}
+
+module "github_token_minter" {
+  source = "git::https://github.com/abcxyz/github-token-minter.git//terraform?ref=f1144811e6f6583b9aa971189ca8f6ef0f947c8a"
+
+  project_id = local.project_id
+
+  domains    = ["REPLACE_DOMAINS"]
```
verbanicm, line 11: Just a note, as part of documentation, we should specify that a domain name will be required

abc.templates/infra/backend.tf
verbanicm: Noticed, there is a separate file, I would merge the backend.tf file and the terraform.tf file into the terraform.tf file.

abc.templates/README.md
```diff
@@ -0,0 +1,18 @@
...
+--input=automation_service_account_member=<SERVICE_ACCOUNT> \
+--input=project_id=<PROJECT_ID> \
+--input=domains=<COMMA_SEPARATED_DOMAINS> \
+--input=terraform_state_bucket=<TERRAFORM_STATE_BUCKET> \
+--input=terraform_state_prefix=<TERRAFORM_STATE_PREFIX> \
+--input=github_owner_id=<GITHUB_ORG_ID> \
+github.com/abcxyz/github-token-minter.git//abc.templates/infra?ref=<SHA>
+```
+
+At least 1 domain is required for GitHub Token Minter.
```
verbanicm, line 18: Just FYI, we have a list of domains for the input to allow domain migrations, I think its safe to say a domain is required. Also, the domains input can be singular for bootstrapping.

abc.templates/README.md
```diff
@@ -0,0 +1,18 @@
...
+```
+abc templates render \
+--dest infra/github-token-minter \
+--input=automation_service_account_member=<SERVICE_ACCOUNT> \
+--input=project_id=<PROJECT_ID> \
+--input=domains=<COMMA_SEPARATED_DOMAINS> \
+--input=terraform_state_bucket=<TERRAFORM_STATE_BUCKET> \
+--input=terraform_state_prefix=<TERRAFORM_STATE_PREFIX> \
+--input=github_owner_id=<GITHUB_ORG_ID> \
+github.com/abcxyz/github-token-minter.git//abc.templates/infra?ref=<SHA>
```
verbanicm, line 15: I think as a best practice, this should be a version, not necessarily a SHA (it could be) but for readability and human understanding installing based on a version tag might be cleaner.

abc.templates/infra/main.tf
```diff
@@ -4,20 +4,13 @@ locals {
 }
 
 module "github_token_minter" {
-  source = "git::https://github.com/abcxyz/github-token-minter.git//terraform?ref=f1144811e6f6583b9aa971189ca8f6ef0f947c8a"
+  source = "git::https://github.com/abcxyz/github-token-minter.git//terraform?ref=ce04ede11817b4b73c1279f9dab5ee3d49c6f7d9"
```
verbanicm, line 7: I wonder if this should be a version tag also, because we will have a chicken v egg problem where we dont know the commit sha for THIS PR. So you would have to submit the changes, then submit another one just to update the sha to the previous commit. Not sure of what the best solution is. just something to be mindful of.

abc.templates/README.md
```diff
@@ -0,0 +1,18 @@
...
+```
```
sethvargo, line 16:
```suggestion

```shell
abc templates render \
  --dest infra/github-token-minter \
  --input=automation_service_account_member=<SERVICE_ACCOUNT> \
  --input=project_id=<PROJECT_ID> \
  --input=domains=<COMMA_SEPARATED_DOMAINS> \
  --input=terraform_state_bucket=<TERRAFORM_STATE_BUCKET> \
  --input=terraform_state_prefix=<TERRAFORM_STATE_PREFIX> \
  --input=github_owner_id=<GITHUB_ORG_ID> \
  github.com/abcxyz/github-token-minter.git//abc.templates/infra?ref=<SHA>
```
```

File: abc.templates/README.md (hunk 12 above)
Comment (sethvargo) [reply abc.templates/README.md:15]: I agree - we could do `TAG_OR_SHA`

File: abc.templates/README.md (hunk 11 above)
Comment (sethvargo) [inline abc.templates/README.md:18]:
```suggestion
At least one domain is required for GitHub Token Minter.
```

https://developers.google.com/style/numbers

abc.templates/infra/spec.yaml
```diff
@@ -0,0 +1,62 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: 'cli.abcxyz.dev/v1alpha1'
+kind: 'Template'
+
+desc: 'Generate the infrastructure necessary to run GitHub Token Minter'
+
+inputs:
+  - name: 'project_id'
```
sethvargo, line 21: I think all of these APIs technically accept a Project ID _or_ Project Number (except WIF, which only supports number). I wonder if we should support both?

abc.templates/infra/terraform.tf
```diff
@@ -0,0 +1,17 @@
...
+  required_version = ">= 1.0"
+
+  backend "gcs" {
+    bucket = "REPLACE_BUCKET_NAME"
+    prefix = "REPLACE_BUCKET_PREFIX"
+  }
+
+  required_providers {
+    google = {
+      version = ">= 4.45"
```
sethvargo, line 11: I wonder if this should be a template input with a default value.

terraform/variables.tf
```diff
@@ -15,6 +15,10 @@
 variable "project_id" {
   description = "The GCP project ID."
   type        = string
+  validation {
+    condition     = length(var.project_id) <= 30
+    error_message = "ERROR: project_id must be <= 30 characters."
+  }
```
sethvargo, line 21: I'm a little concerned with this:

1. I find client-side validation to be artificially limiting. If GCP ever changed their project IDs to be longer, template users would still be stuck with the old values, and we increase our long-term maintenance. It's legacy, but project IDs can also be prefixed with their org as a syntax that most of the API endpoints accept (e.g. `mydomain.com:my-project`). I'd prefer if we pushed this validation up into the abc CLI, where users have the ability to bypass a validation easily.

2. There are other validations to consider too (must start with a letter, can only contain lowercase letters, digits, and hyphens, cannot end with a hyphen). I'm curious why we're _only_ validating length? Like I said above, I'd strongly prefer if this validation was pushed into the abc cli where we can both (a) centralize the validation and (b) have a good way for power users to bypass the validation.

abc.templates/infra/terraform.tf
verbanicm: I dont know if default values are supported yet

terraform/variables.tf
verbanicm: Hmmmmmm, pushing to abc CLI makes most sense, assuming validation is implemented already

gjonathanhong: I think input validation for abc CLI is [in-progress](https://github.com/abcxyz/abc/issues/39#issue-1766105040). Would you prefer I keep or remove the validation in terraform now, until input validation with the CLI is available?

abc.templates/README.md
gjonathanhong: ack, thanks for context

terraform/variables.tf
verbanicm: I had a sync with dave today, he said its done, maybe sync with him? Rather do it once.

abc.templates/README.md
```diff
@@ -0,0 +1,18 @@
...
+1. Render infrastructure template and actuate terraform.
+```shell
+abc templates render \
+--dest infra/github-token-minter \
+--input=automation_service_account_member=<SERVICE_ACCOUNT> \
+--input=project_id=<PROJECT_ID> \
+--input=domain=<DOMAIN> \
+--input=terraform_state_bucket=<TERRAFORM_STATE_BUCKET> \
+--input=terraform_state_prefix=<TERRAFORM_STATE_PREFIX> \
+--input=github_owner_id=<GITHUB_ORG_ID> \
```
verbanicm, line 14: I believe per Seths previous comment these input flags should be single dash (-) not double (--)

abc.templates/infra/spec.yaml
verbanicm: @sethvargo

sethvargo: Sticking with just project ID is fine.

abc.templates/README.md
gjonathanhong: got it, https://github.com/abcxyz/abc#flags will have to be updated as well

abc.templates/infra/main.tf
gjonathanhong: I changed this to v.0.0.11 so that when we push a v0.0.11 tag it will match.

verbanicm (approved):

PR #57, opened 2023-09-05
feat: add deployments template for bootstrapping GTM with abc CLI

gjonathanhong:


bradegler (dismissed):

abc.templates/deployments/contents/configs/README.md
```diff
@@ -0,0 +1,65 @@
...
+  if: 'assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml")'
+  repositories:
+    - 'github-token-minter'
+  permissions:
+    issues: 'read'
+```
+
+The `if` clause uses the [Google Common Expression Language](https://github.com/google/cel-spec) to match an inbound OIDC token against a series of rules. Any attributes from the assertion object are available to match against and you can make this expression as simple or complicated as required.
+
+The object mirrors what is available in an OIDC token for a GitHub Workflow, below is an example token.
```
verbanicm, line 28: I would add a link, as this changes over time: 
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#configuring-the-oidc-trust-with-the-cloud

verbanicm (approved):

sailorlqh (approved):

PR #58, opened 2023-09-05
feat: add deployment workflows to the deployments template

gjonathanhong:
By default, the deployments template will generate the following structure to build and deploy github-token-minter.
```
abcxyz-services
   .github/workflows/
         deploy-github-token-minter.yaml
   deployments/github-token-minter/
         configs/
             README.md
         deploy.sh
         Dockerfile
```


abc.templates/deployments/contents/workflows/deploy-github-token-minter.yaml
```diff
@@ -0,0 +1,53 @@
...
+        uses: 'google-github-actions/auth@35b0e87d162680511bf346c299f71c9c5c379033' # ratchet:google-github-actions/auth@v1
+        with:
+          workload_identity_provider: '${{ env.WIF_PROVIDER }}'
+          service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
+          token_format: 'access_token'
+
+      - name: 'Setup gcloud'
+        uses: 'google-github-actions/setup-gcloud@e30db14379863a8c79331b04a9969f4c1e225e0b' # ratchet:google-github-actions/setup-gcloud@v1
+
+      - uses: 'docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc' # ratchet:docker/login-action@v2
```
sethvargo, line 42: Something to consider here: the authentication token is usually only valid for a few minutes (5min by default). If the build step takes more than 5 minutes, then the auth token provided to `docker login` will have expired and the push will fail. It'd be more full-proof to do something like:

- auth
- docker login (so `FROM` tags can use private registries)
- docker build
- auth (again, get a new token)
- docker login (again with the new token)
- docker push

gjonathanhong: That is a valid point but I expect a short build time for bootstrapping because we're only adding configuration files on top of the parent image. The intention of having everything in `deploy.sh` is to more easily identify the version of github-token-minter, but if the build step is a concern then we should separate it accordingly to your full-proof solution.

verbanicm: The idea here was for readability, the source of truth is the deploy script. Lets keep this as-is and monitor build times. I dont see this surpassing 5 min, at least in the near term.

abc.templates/deployments/spec.yaml
```diff
@@ -21,23 +21,55 @@ inputs:
...
+  - name: 'service_name'
+    desc: 'The name of the GitHub Token Minter Cloud Run service'
 
 steps:
   - desc: 'Include required Dockerfiles and configs'
     action: 'include'
     params:
       paths:
         - 'contents'
+        - 'contents/workflows'
```
verbanicm, line 49: Should these be siblings instead? It seems strange that contents includes all children but then workflows is "de-referenced".

abc.templates/deployments/contents/deploy.sh
```diff
@@ -0,0 +1,10 @@
+IMAGE_NAME=REPLACE_FULL_IMAGE_NAME:REPLACE_VERSION
```
verbanicm, line 1: This needs to be executable and a script.

Add `#!/usr/bin/env bash`
Add `set -euo pipefail`

verbanicm (approved):

PR #78, opened 2023-10-25
chore: use subdirectory input for output and bucket prefix of infra template

gjonathanhong:


abc.templates/infra/spec.yaml
```diff
@@ -60,7 +58,7 @@ steps:
         - to_replace: 'REPLACE_BUCKET_NAME'
           with: '{{.terraform_state_bucket}}'
         - to_replace: 'REPLACE_BUCKET_PREFIX'
-          with: '{{.terraform_state_prefix}}'
+          with: '{{.subdirectory}}'
```
verbanicm, line 61: Should we use dest instead? What if someone uses dest and this flag?

verbanicm: Lets merge this for now and re-work if needed

verbanicm (approved):

abc.templates/infra/spec.yaml
gjonathanhong: Will address this in following changes

PR #80, opened 2023-10-25
chore: use custom name to customize cloud run resource names and log sink name

gjonathanhong:
Uses `custom_name` to override the default values of cloud run and log sink resource names. This also adds validation rules on the `custom_name` input variable.

### DIFF
--- a/abc.templates/infra/spec.yaml
+++ b/abc.templates/infra/spec.yaml
@@ -21,6 +21,11 @@ inputs:
   - name: 'custom_name'
     desc: 'A custom name for Github Token Minter'
     default: 'github-token-minter'
+    rules:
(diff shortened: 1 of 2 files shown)


abc.templates/infra/spec.yaml
```diff
@@ -21,6 +21,11 @@ inputs:
   - name: 'custom_name'
     desc: 'A custom name for Github Token Minter'
     default: 'github-token-minter'
+    rules:
+      - rule: 'size(custom_name) <= int(19)'
```
gjonathanhong, line 25: This is the longest allowable length before reaching the first GCP resource name limitation.

bradegler (approved):

PR #81, opened 2023-10-25
chore: use name input variable for naming resources

gjonathanhong:


bradegler (approved):

terraform/github-wif.tf
```diff
@@ -25,9 +25,9 @@ locals {
 resource "google_iam_workload_identity_pool" "default" {
   project = var.project_id
 
-  workload_identity_pool_id = "github-token-minter"          # 32 characters
-  display_name              = "GitHub Token Minter WIF pool" # 32 characters
-  description               = "GitHub Token Minter OIDC identity pool"
+  workload_identity_pool_id = var.name               # 32 characters
```
verbanicm, line 28: Are we/should we do any validation on the name variable? e.g. only alphanumeric and hyphens?

gjonathanhong: Yes, the name variable has existing validation here https://github.com/abcxyz/github-token-minter/blob/main/terraform/variables.tf#L25

gjonathanhong: For the `custom_name` input variable of the templates, I think we should add validation for the max length of the string but using the CLI to perform validations

verbanicm: That makes sense.

verbanicm (approved):

PR #107, opened 2024-02-27
fix: add audit logging back

### LINKED ISSUE #103: Add audit log functionality back
### TL;DR

Audit logging via Lumberjack was removed due to bug. Re-add audit logging functionality, without lumberjack dependency (if possible).

### Expected behavior

_No response_

### Observed behavior

_No response_

### Environment Details

```markdown
All
```

### Additional information

_No response_

verbanicm:
Fixes #103

### DIFF
--- a/pkg/server/token_minter.go
+++ b/pkg/server/token_minter.go
@@ -25,7 +25,6 @@ import (
 	"io"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/lestrrat-go/jwx/v2/jwt"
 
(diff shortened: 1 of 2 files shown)


pkg/server/token_minter.go
```diff
@@ -109,13 +112,69 @@ func (s *TokenMintServer) handleToken() http.Handler {
...
+		Request: &structpb.Struct{Fields: map[string]*structpb.Value{
+			"token":   structpb.NewStringValue(string(token)),
+			"request": structpb.NewStringValue(string(req)),
+			"config":  structpb.NewStringValue(string(config)),
+		}},
+		AuthenticationInfo: &audit.AuthenticationInfo{
+			PrincipalEmail: principal,
+		},
+	}
+
```
sethvargo, line 172: Is there a reason we need to use the Audit Logging proto? This is just a standard JSON log, so we can use our own format. I'd prefer to avoid bringing in grpc and all these deps.

I think we really just want something like:

```go
logger.DebugContext(ctx, "generating token", 
  "request.token": token, 
  "request.config": config,
  "resource": resourceName,
  "principal": principal)
```

verbanicm: Ack

pkg/server/token_minter.go
```diff
@@ -218,6 +199,12 @@ func (s *TokenMintServer) processRequest(r *http.Request, auditEvent *auditEvent
 	// Replace the requested repository list with actual values
 	request.Repositories = repos
 
+	logger.InfoContext(ctx, "generating token",
```
verbanicm, line 202: @sethvargo will the logger print this correctly or should I turn this into JSON?

verbanicm: tested, this works appropriately

PR #119, opened 2024-07-22
feat: multi-tenacy implementation

bradegler:
Updated to lookup the proper GitHub App installation based on the organization name and the repository name instead of using the configuration value provided on the command line.

### DIFF
--- a/pkg/server/token_minter.go
+++ b/pkg/server/token_minter.go
@@ -25,12 +25,10 @@ import (
 	"io"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/lestrrat-go/jwx/v2/jwt"
 
 	"github.com/abcxyz/github-token-minter/pkg/version"
-	"github.com/abcxyz/pkg/cache"
 	"github.com/abcxyz/pkg/gcputil"
 	"github.com/abcxyz/pkg/githubauth"
 	"github.com/abcxyz/pkg/logging"
@@ -44,10 +42,9 @@ const (
 // TokenMintServer is the implementation of an HTTP server that exchanges
 // a GitHub OIDC token for a GitHub application token with eleveated privlidges.
 type TokenMintServer struct {
-	githubApp         *githubauth.App
-	configStore       ConfigReader
-	jwtParseOptions   []jwt.ParseOption
-	githubAppInstalls *cache.Cache[*githubauth.AppInstallation]
+	githubApp       *githubauth.App
+	configStore     ConfigReader
+	jwtParseOptions []jwt.ParseOption
 }
 
 type oidcClaims struct {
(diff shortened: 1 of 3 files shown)


pkg/server/token_minter.go
```diff
@@ -72,11 +75,12 @@ type oidcClaims struct {
...
-func NewRouter(ctx context.Context, githubAppInstallation *githubauth.AppInstallation, configStore ConfigReader, jwtParseOptions []jwt.ParseOption) (*TokenMintServer, error) {
+func NewRouter(ctx context.Context, githubApp *githubauth.App, configStore ConfigReader, jwtParseOptions []jwt.ParseOption) (*TokenMintServer, error) {
 	return &TokenMintServer{
-		githubAppInstallation: githubAppInstallation,
-		configStore:           configStore,
-		jwtParseOptions:       jwtParseOptions,
+		githubApp:         githubApp,
+		configStore:       configStore,
+		jwtParseOptions:   jwtParseOptions,
+		githubAppInstalls: cache.New[*githubauth.AppInstallation](1 * time.Hour),
```
sethvargo, line 83: You actually don't need to cache these; they are cached automatically by the library.

bradegler: Oh, great. I'll remove that then.

verbanicm (approved):

pkg/server/token_minter.go
```diff
@@ -177,12 +177,25 @@ func (s *TokenMintServer) processRequest(r *http.Request) (int, string, error) {
...
 
+	// The repository claim is of the form <org_name>/<repo_name>.
+	// Use this string split instead of attempting to use this and the repository_owner claim since
+	// the repository_owner claim is optional.
+	repoParts := strings.Split(claims.Repository, "/")
+	if len(repoParts) != 2 {
+		return http.StatusBadRequest, fmt.Sprintf("'repository' claim formatted incorrectly, requires <org_name>/<repo_name> format - received [%s]", claims.Repository), nil
+	}
+	// Lookup the App installation for the GitHub owner/repo
+	installation, err := s.githubApp.InstallationForRepo(ctx, repoParts[0], repoParts[1])
```
yolocs, line 188: I wonder what happens if the installation is in the org rather than the repo?

pkg/server/token_minter.go
yolocs: I'm curious - the cache in pkg https://github.com/abcxyz/pkg/blob/main/githubauth/app.go#L268-L282 seems to only cache "installations" but it seems we want to cache access tokens (so we can reuse them before timeout)?

pkg/server/token_minter.go
bradegler: The installation is at the org level, but when you configure it you either configure it for `all repositories` or specific individual repositories. There is no org without repos option AFAIK, I guess you could select no repos but then it would be useless.

pkg/server/token_minter.go
sethvargo: > I'm curious - the cache in pkg https://github.com/abcxyz/pkg/blob/main/githubauth/app.go#L268-L282 seems to only cache "installations" but it seems we want to cache access tokens (so we can reuse them before timeout)?

Correct - the githubapp package caches the lookup of an installation ID from inputs. It does not cache any generated tokens.

bradegler: Caching the access token would be problematic since each access token might have a different set of permissions it is requesting. We'd need to devise a cache key that encapsulated the set of permissions or something. Its better to request a fresh access token each time rather than risk using the wrong token and giving more permissions than required.

PR #122, opened 2024-09-06
feat: initial implementation of configuration file updates for v2

bradegler:
After playing with it a bit, I feel like this is how I want configuration files to be structured and function.

NOTE: I have not integrated this into the actual request flow yet, looking for feedback on the design first before doing that work.

Example: 

```
version: 'minty.abcxyz.dev/v2'
# Top level rule that is evaluated first. All scopes basically inherit this when we validate the request
rule:
  if: |-
      assertion.iss = 'https://token.actions.githubusercontent.com' &&
      assertion.organization_id = '93787867' &&
      assertion.repository_id = '576289489' &&
      assertion.ref != 'refs/heads/main'

# Request level scopes that can be targeted, each scope provides a set of permissions that are available.
# The scope is looked up by name and then the validation in the "if" is applied to ensure that it matches
# the expected criteria.
scope:
  draft-release:
    rule:
      if: |-
        assertion.workflow_ref == assertion.job_workflow_ref &&
        assertion.job_workflow_ref == 'abcxyz/github-token-minter/.github/workflows/draft-release.yml@refs/heads/main' &&
        assertion.event_name == 'workflow_dispatch'
    permissions:
      contents: 'write'
      pull_requests: 'write'

  release:
    rule:
      if: |-
        assertion.workflow_ref == assertion.job_workflow_ref &&
        assertion.job_workflow_ref == 'abcxyz/github-token-minter/.github/workflows/release.yml@refs/heads/main' &&
        assertion.event_name == 'push'
    permissions:
      contents: 'write'`
```


yolocs:
LGTM the design. One thing we might want to consider is to versionize the Config. Since in v2 the Config could live in other people's repo and managed by the corresponding org/repo owners, it's hard to say that we will never need to make breaking changes in the Config structure.

verbanicm:
@bradegler Are you expecting that scope is a required parameter or can I leave it empty and get some default permissions? It looks like scope should be required but wanted to confirm (IMO, scope should be required).

bradegler:
> LGTM the design. One thing we might want to consider is to versionize the Config. Since in v2 the Config could live in other people's repo and managed by the corresponding org/repo owners, it's hard to say that we will never need to make breaking changes in the Config structure.

That's a good suggestion. I'll add a version to the doc but I'm not going to worry about handling multiple versions until we have a second version.

bradegler:
> @bradegler Are you expecting that scope is a required parameter or can I leave it empty and get some default permissions? It looks like scope should be required but wanted to confirm (IMO, scope should be required).

It is required and I'll handle that on the API side when I integrate these changes.

pkg/server/config/config_store.go
```diff
@@ -0,0 +1,149 @@
...
+	localLoader := localConfigFileLoader{configDir: localConfigDir}
+	inRepoLoader := ghInRepoConfigFileLoader{
+		provider: func(ctx context.Context, org, repo string) (*github.Client, error) {
+			return makeGitHubClient(ctx, app, org, repo)
+		},
+		configPath: ".github/minty.yaml",
+		ref:        "main",
+	}
+	orgLoader := fixedRepoConfigFileLoader{
+		repo: ".google-github",
```
verbanicm, line 61: Should this be configurable given this an oss project?

verbanicm (approved):

pkg/server/config/config_store.go
bradegler: Yes, that will be passed in when I integrate this with the rest of the code. I just hard coded it for now so you knew what the intended values are.

yolocs (approved):

PR #127, opened 2024-09-24
feat: converted command into cli format in preparation for more commands

bradegler:


pkg/cli/root.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+
+	"github.com/abcxyz/github-token-minter/pkg/version"
+	"github.com/abcxyz/pkg/cli"
+)
+
+var rootCmd = func() cli.Command {
+	return &cli.RootCommand{
+		Name:    "github-token-minter",
```
verbanicm, line 27: minty?

verbanicm (approved):

pkg/cli/root.go
bradegler: Good call

sethvargo (approved):

PR #132, opened 2024-09-27
fix: removing cache to accelerate testing cycle

bradegler:


verbanicm (approved):

verbanicm (approved):

pkg/config/config_evaluator.go
```diff
@@ -46,24 +46,43 @@ func NewConfigEvaluator(expireAt time.Duration, localConfigDir, repoConfigPath,
 	if err != nil {
 		return nil, fmt.Errorf("failed to create CEL environment: %w", err)
 	}
+	enableCaching := false
```
sethvargo, line 49: Drive-by. I would prefer if we had a runtime config for "cache duration" and we could set it to something super small like "1s"

sethvargo: @bradegler ^

bradegler: Yes. It will get moved into a command line option. I have that already teed up in a follow up.

PR #135, opened 2024-09-30
feat: updated minty config

bradegler:


.github/minty.yaml
```diff
@@ -1,8 +1,24 @@
 version: 'minty.abcxyz.dev/v2'
+
 rule:
-  if: true
```
sethvargo, line 3: I'm curious why this changed?

sethvargo (approved):

.github/minty.yaml
bradegler: The original config was built just for integration testing. 

The latest config covers integ and the create-tag workflow as it was defined in the `abcxyz-services` repo.

Issue #134, opened 2024-09-30
Create CEL variables for well known issuers


verbanicm:
### TL;DR

Prevent mistakes by providing CEL variables for well known OIDC token issuers.

### Detailed design

Instead of:
`assertion.iss == 'https://token.actions.githubusercontent.com'`

Add variables so the syntax becomes:

`assertion.iss == issuers.github`

bradegler:
@Celdir - has this issue been completed?

Celdir:
Yep, this is done. Closing now

PR #139, opened 2024-10-01
Add CEL variable for well known issuers

Celdir:
Prevent mistakes by providing CEL variable for well known OIDC token issuers.

https://github.com/abcxyz/github-token-minter/issues/134

### DIFF
--- a/.github/minty.yaml
+++ b/.github/minty.yaml
@@ -2,7 +2,7 @@ version: 'minty.abcxyz.dev/v2'
 
 rule:
   if: |-
-    assertion.iss == 'https://token.actions.githubusercontent.com' &&
+    assertion.iss == issuers.github &&
     assertion.repository_owner_id == '93787867' &&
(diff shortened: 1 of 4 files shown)


.github/minty.yaml
```diff
@@ -2,7 +2,7 @@ version: 'minty.abcxyz.dev/v2'
 
 rule:
   if: |-
-    assertion.iss == 'https://token.actions.githubusercontent.com' &&
+    assertion.iss == issuers.github &&
```
Celdir, line 5: @drevell is it safe to change this now or do I need to wait for minty to deploy before updating its own config?

drevell: Better ask Brad.

verbanicm: you will need to redeploy.

verbanicm (changes requested):
change minty config after deployment

verbanicm (approved):

PR #141, opened 2024-10-03
feat: additional diagnostics to make debugging easier in prod

bradegler:
* Fixed a few configuration options that weren't being used
* Adjusted logging levels for startup configuration details
* Added additional logging in the config loaders to output the path to where the configuration is coming from e.g. `https://github.com/abcxyz/github-token-minter/.github/minty.yaml`. This will help when a scope doesn't resolve so we are sure its looking in the correct files.

### DIFF
--- a/pkg/config/config_evaluator.go
+++ b/pkg/config/config_evaluator.go
@@ -36,6 +36,11 @@ const (
 type GitHubClientProvider func(ctx context.Context, org, repo string) (*github.Client, error)
 
 type ConfigEvaluator interface {
+	// Eval looks for a configuration for the combination of org and repo that contains
+	// the requested scope. If a match is made the scope is returned. This function also
+	// returns a string that represents the path to where the scope came from or the file that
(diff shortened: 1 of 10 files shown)


pkg/config/config_evaluator.go
```diff
@@ -36,14 +36,14 @@ const (
 type GitHubClientProvider func(ctx context.Context, org, repo string) (*github.Client, error)
 
 type ConfigEvaluator interface {
-	Eval(ctx context.Context, org, repo, scope string, token interface{}) (*Scope, error)
+	Eval(ctx context.Context, org, repo, scope string, token interface{}) (*Scope, string, error)
```
drevell, line 39: Perhaps drop a comment here explaining what the returned string is

drevell (approved):

pkg/config/config_evaluator.go
bradegler: done

PR #144, opened 2024-10-07
feat: add verification for non-github oidc tokens

Celdir:
* Dynamically resolve JWK key set used for token verification based on token's issuer URL and OpenID Configuration and cache key sets in the same way that we do in [JVS](https://github.com/abcxyz/jvs/blob/main/apis/v0/client.go#L36-L39). 
* Cache `jwks_uri` for issuers with a command line configured (default 4 hours) TTL (these should very rarely change) so we don't have to GET the OpenID Configs often. 
* Command line configured allowlist of issuers, defaulting to just GitHub and Google

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -16,7 +16,9 @@ package server
 
 import (
 	"fmt"
+	"strings"
 
+	"github.com/abcxyz/github-token-minter/pkg/config"
 	"github.com/abcxyz/pkg/cli"
 )
 
@@ -24,7 +26,6 @@ type Config struct {
 	Port       string
 	AppID      string
 	PrivateKey string
-	JWKSUrl    string
 
 	// GitHubAPIBaseURL is the base URL for the GitHub installation. It should
 	// include the protocol (https://) and no trailing slashes.
@@ -36,6 +37,9 @@ type Config struct {
 	OrgConfigPath      string
 	Ref                string
 	ConfigCacheSeconds string
+
+	JWKSURICacheSeconds string
+	IssuerAllowlist     string
 }
 
 // Validate validates the artifacts config after load.
@@ -45,9 +49,6 @@ func (cfg *Config) Validate() error {
 	}
 	if cfg.PrivateKey == "" {
 		return fmt.Errorf("GITHUB_PRIVATE_KEY is required")
-	}
-	if cfg.JWKSUrl == "" {
-		cfg.JWKSUrl = "https://token.actions.githubusercontent.com/.well-known/jwks"
 	}
 	if cfg.ConfigCacheSeconds == "" {
 		cfg.ConfigCacheSeconds = "900"
@@ -64,6 +65,14 @@ func (cfg *Config) Validate() error {
 	}
 	if cfg.Ref == "" {
 		cfg.Ref = "main"
+	}
+
+	if cfg.JWKSURICacheSeconds == "" {
+		// 4 hours default
+		cfg.JWKSURICacheSeconds = "14400"
+	}
+	if cfg.IssuerAllowlist == "" {
+		cfg.IssuerAllowlist = strings.Join([]string{config.GitHubIssuer, config.GoogleIssuer}, ",")
 	}
 
 	return nil
@@ -141,6 +150,20 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 		Target: &cfg.ConfigCacheSeconds,
 		EnvVar: "CONFIG_CACHE_MINUTES",
 		Usage:  `The number of minutes to cache configuration files before retrieving fresh ones. Defaults to 15 minutes.`,
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:   "jwks-uri-cache-seconds",
+		Target: &cfg.JWKSURICacheSeconds,
+		EnvVar: "JWKS_URI_CACHE_SECONDS",
+		Usage:  `The number of seconds to cache the jwks_uri from an OIDC token issuer's OpenID Configuration before retrieving fresh ones. Defaults to 4 hours.`,
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:   "issuer-allowlist",
+		Target: &cfg.IssuerAllowlist,
+		EnvVar: "ISSUER_ALLOWLIST",
+		Usage:  `The list of OIDC token issuers that GitHub Token Minter will accept. Format is a comma-separated list of URLs. Defaults to GitHub and Google issuers.`,
 	})
 
 	return set
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/abcxyz/github-token-minter/pkg/config"
 	"github.com/abcxyz/pkg/cli"
 )
 
@@ -147,7 +148,7 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 		Name:    "jwks-cache-duration",
 		Target:  &cfg.JWKSCacheDuration,
 		EnvVar:  "JWKS_CACHE_DURATION",
-		Usage:   `The duration to cache the JWKS for an OIDC token issuer. Defaults to 4 hours.`,
+		Usage:   `The duration for which to cache the JWKS for an OIDC token issuer.`,
 		Default: 4 * time.Hour,
 	})
 
@@ -155,7 +156,7 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 		Name:    "issuer-allowlist",
 		Target:  &cfg.IssuerAllowlist,
 		EnvVar:  "ISSUER_ALLOWLIST",
-		Usage:   `The list of OIDC token issuers that GitHub Token Minter will accept. Format is a comma-separated list of URLs. Defaults to GitHub and Google issuers.`,
+		Usage:   `The list of OIDC token issuers that GitHub Token Minter will accept. Format is a comma-separated list of URLs or the flag can be specified multiple times.`,
 		Default: []string{config.GitHubIssuer, config.GoogleIssuer},
 	})
 
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -144,17 +144,19 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 	})
 
 	f.DurationVar(&cli.DurationVar{
-		Name:   "jwks-cache-duration",
-		Target: &cfg.JWKSCacheDuration,
-		EnvVar: "JWKS_CACHE_DURATION",
-		Usage:  `The duration to cache the JWKS for an OIDC token issuer. Defaults to 4 hours.`,
+		Name:    "jwks-cache-duration",
+		Target:  &cfg.JWKSCacheDuration,
+		EnvVar:  "JWKS_CACHE_DURATION",
+		Usage:   `The duration to cache the JWKS for an OIDC token issuer. Defaults to 4 hours.`,
+		Default: 4 * time.Hour,
 	})
 
 	f.StringSliceVar(&cli.StringSliceVar{
-		Name:   "issuer-allowlist",
-		Target: &cfg.IssuerAllowlist,
-		EnvVar: "ISSUER_ALLOWLIST",
-		Usage:  `The list of OIDC token issuers that GitHub Token Minter will accept. Format is a comma-separated list of URLs. Defaults to GitHub and Google issuers.`,
+		Name:    "issuer-allowlist",
+		Target:  &cfg.IssuerAllowlist,
+		EnvVar:  "ISSUER_ALLOWLIST",
+		Usage:   `The list of OIDC token issuers that GitHub Token Minter will accept. Format is a comma-separated list of URLs. Defaults to GitHub and Google issuers.`,
+		Default: []string{config.GitHubIssuer, config.GoogleIssuer},
 	})
 
 	return set
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -16,9 +16,8 @@ package server
 
 import (
 	"fmt"
-	"strings"
+	"time"
 
-	"github.com/abcxyz/github-token-minter/pkg/config"
 	"github.com/abcxyz/pkg/cli"
 )
 
@@ -38,8 +37,8 @@ type Config struct {
 	Ref                string
 	ConfigCacheSeconds string
 
-	JWKSURICacheSeconds string
-	IssuerAllowlist     string
+	JWKSURICacheDuration time.Duration
+	IssuerAllowlist      []string
 }
 
 // Validate validates the artifacts config after load.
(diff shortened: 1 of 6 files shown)


pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+
+func NewOIDCResolver(ctx context.Context) *OIDCResolver {
+	return &OIDCResolver{
+		cache: jwk.NewCache(ctx),
+		// 24 hour cache timeout, jwks_uri should change very infrequently
+		issuerToJwksURI: cache.New[string](time.Hour * 24),
+	}
+}
+
+func (r *OIDCResolver) ResolveKeySet(ctx context.Context, oidcHeader string) (jwk.Set, error) {
```
sethvargo, line 53: Doc

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+	issuer, err := r.extractIssuer(ctx, oidcHeader)
```
sethvargo, line 54: Why is this a helper? It seems to be adding an unnecessary layer of indirection.

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract issuer from token: %w", err)
+	}
+
+	jwksURI, err := r.registerIssuer(ctx, issuer)
```
sethvargo, line 59: Similarly, why is this a helper? It seems to be adding an unnecessary layer of indirection.

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+		if err != nil {
+			return "", fmt.Errorf("failed to register JWKS URI %q to cache: %w", jwksURI, err)
+		}
+	}
+
+	return jwksURI, nil
+}
+
+func (r *OIDCResolver) resolveJwksURI(ctx context.Context, issuer string) (string, error) {
+	uri, cached := r.issuerToJwksURI.Lookup(issuer)
```
sethvargo, line 93: This is a data race, and I'm also not sure it's necessary to cache this.

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (r *OIDCResolver) registerIssuer(ctx context.Context, issuer string) (string, error) {
+	jwksURI, err := r.resolveJwksURI(ctx, issuer)
+	if err != nil {
+		return "", fmt.Errorf("failed to resolve JWKS URI for %q: %w", issuer, err)
+	}
+
+	if !r.cache.IsRegistered(jwksURI) {
+		err = r.cache.Register(jwksURI)
```
sethvargo, line 83: This is a data race.

sethvargo, line 83: > Register does not check for the validity of the url being registered. If you need to make sure that a url is valid before entering your main loop, call `Refresh` once to make sure the JWKS is available.

Source: https://pkg.go.dev/github.com/lestrrat-go/jwx/v2/jwk#Cache.Register

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return "", fmt.Errorf("error creating GET request for %q: %w", configURL, err)
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("error fetching OpenID Configuration from %q: %w", configURL, err)
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
```
sethvargo, line 113: Limit the max response size to something like 24MiB just to be safe.

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to read response body from GET %q: %w", configURL, err)
+	}
+
+	var config OpenIDConfiguration
+	err = json.Unmarshal(body, &config)
+	if err != nil {
+		return "", fmt.Errorf("failed to unmarshal OpenID Configuration: %w", err)
+	}
```
sethvargo, line 122: Cuddle errors

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (r *OIDCResolver) registerIssuer(ctx context.Context, issuer string) (string, error) {
+	jwksURI, err := r.resolveJwksURI(ctx, issuer)
+	if err != nil {
+		return "", fmt.Errorf("failed to resolve JWKS URI for %q: %w", issuer, err)
+	}
+
+	if !r.cache.IsRegistered(jwksURI) {
+		err = r.cache.Register(jwksURI)
+		if err != nil {
```
sethvargo, line 84: Cuddle errors

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+	jwksURI, err := r.registerIssuer(ctx, issuer)
+	if err != nil {
+		return nil, fmt.Errorf("failed to register issuer %q: %w", issuer, err)
+	}
+
+	return jwk.NewCachedSet(r.cache, jwksURI), nil
+}
+
+func (r *OIDCResolver) extractIssuer(ctx context.Context, oidcHeader string) (string, error) {
+	// Parse without validation to extract issuer so we can use it to obtain the key set to use for validation
```
sethvargo, line 68:
```suggestion
	// Parse without verification to extract issuer so we can use it to obtain the key set to use for verification
```

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+type OpenIDConfiguration struct {
+	// The only field we need from the config
+	JwksURI string `json:"jwks_uri"`
+}
+
+func NewOIDCResolver(ctx context.Context) *OIDCResolver {
+	return &OIDCResolver{
+		cache: jwk.NewCache(ctx),
+		// 24 hour cache timeout, jwks_uri should change very infrequently
+		issuerToJwksURI: cache.New[string](time.Hour * 24),
```
sethvargo, line 49: I recommend settings this to something smaller like 4 hours for now.

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+func (r *OIDCResolver) extractIssuer(ctx context.Context, oidcHeader string) (string, error) {
+	// Parse without validation to extract issuer so we can use it to obtain the key set to use for validation
+	token, err := jwt.ParseString(oidcHeader, jwt.WithContext(ctx), jwt.WithVerify(false))
+	if err != nil {
+		return "", fmt.Errorf("failed to parse jwt header: %w", err)
+	}
+	return token.Issuer(), nil
+}
+
+func (r *OIDCResolver) registerIssuer(ctx context.Context, issuer string) (string, error) {
```
sethvargo, line 76: Should we limited the allowed issuers? As written, someone could DOS the service with millions of issuers that we would never actually accept, and each request would result in us making upstream requests to other people's servers. It feels like we want a list of allowed issuers as a runtime configuration.

sethvargo (changes requested):
There are a number of data racers in this PR, probably more than I pointed out because there are too many helper functions and layers of indirection.

I'm also concerned about accepting literally any issuer. It seems like we want to maintain an allowlist of issuers as a runtime configuration.

pkg/server/jwk_resolver.go
bradegler: Can we make this configurable too from the command line with a default?

pkg/server/jwk_resolver.go
Celdir: Removed this helper

pkg/server/jwk_resolver.go
Celdir: Removed this helper, but added a different one only because I'm using a mutex to guard cache registers now and wanted to use `defer mu.Unlock()`

pkg/server/jwk_resolver.go
Celdir: Resolved by using `WriteThruLookup` instead

pkg/server/jwk_resolver.go
Celdir: Resolved by using mutex to guard cache registers

pkg/server/jwk_resolver.go
Celdir: Done

pkg/server/jwk_resolver.go
Celdir: Added a config option from command line from this that defaults to 4 hours

pkg/server/jwk_resolver.go
Celdir: Added an allowlist of issuers configured from the command line, defaulting to just GitHub and Google issuers

pkg/server/jwk_resolver.go
Celdir: Done

pkg/server/jwk_resolver.go
Celdir: Done

pkg/server/jwk_resolver.go
Celdir: Done

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	if err := r.cache.Register(jwksURI); err != nil {
+		return fmt.Errorf("failed to register JWKS URI %q to cache: %w", jwksURI, err)
+	}
+	// call Refresh to validate URI
+	if _, err := r.cache.Refresh(ctx, jwksURI); err != nil {
+		return fmt.Errorf("failed to refresh JWKS URI %q in cache: %w", jwksURI, err)
+	}
+	return nil
+}
```
sethvargo, line 100: Sorry, but I'm still really confused by these layers of indirection. We've got:

```text
issuer -> jwks_url  -> jwks
```

and we're caching both `issuer -> jwks_url` **and** `jwks_url` -> `jwks`. It seems like we actually want to cache `issuer` -> `jwks` for some reasonable/configurable period of time (e.g. 4 hours).

pkg/server/config.go
```diff
@@ -160,5 +169,19 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 		Usage:  `The number of minutes to cache configuration files before retrieving fresh ones. Defaults to 15 minutes.`,
 	})
 
+	f.StringVar(&cli.StringVar{
```
sethvargo, line 172: Use the time.Duration instead (that way you're not limited to seconds)

pkg/server/config.go
```diff
@@ -40,6 +41,9 @@ type Config struct {
 	OrgConfigPath      string `env:"ORG_CONFIG_PATH,default=minty.yaml"`
 	Ref                string `env:"REF,default=main"`
 	ConfigCacheSeconds string `env:"CONFIG_CACHE_SECONDS=900"`
+
+	JWKSURICacheSeconds string `env:"JWKS_URI_CACHE_SECONDS=14400"`
```
sethvargo, line 45: time.Duration instead

pkg/server/config.go
```diff
@@ -70,6 +71,14 @@ func (cfg *Config) Validate() error {
 		cfg.Ref = "main"
 	}
 
+	if cfg.JWKSURICacheSeconds == "" {
+		// 4 hours default
+		cfg.JWKSURICacheSeconds = "14400"
+	}
+	if cfg.IssuerAllowlist == "" {
+		cfg.IssuerAllowlist = strings.Join([]string{config.GitHubIssuer, config.GoogleIssuer}, ",")
+	}
```
sethvargo, line 80: You can declare these defaults with envconfig or in the CLI; it's a bit strange to do it here.

Celdir: Removed this section

pkg/server/config.go
Celdir: Done

pkg/server/config.go
Celdir: Done

pkg/server/jwk_resolver.go
```diff
@@ -0,0 +1,129 @@
...
+// ResolveKeySet finds and caches the correct OIDC key set to use based on the token's issuer's OpenID Configuration.
+func (r *OIDCResolver) ResolveKeySet(ctx context.Context, oidcHeader string) (jwk.Set, error) {
+	// Parse without verification to extract issuer so we can use it to obtain the key set to use for verification
+	token, err := jwt.ParseString(oidcHeader, jwt.WithContext(ctx), jwt.WithVerify(false), jwt.WithValidate(false))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse issuer from jwt header: %w", err)
+	}
+	issuer := token.Issuer()
+
+	if !slices.Contains(r.issuerAllowlist, issuer) {
```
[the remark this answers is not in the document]
Celdir: Used a string slice, mainly because there's no flag convenience function for `map[string]struct{}` and `map[string]string` seems counterintuitive to me. Do you think we should go and add a convenience function for `StringSetVar` that resolves to `map[string]struct{}`? I can do that if you think it would be better.

pkg/server/config.go
```diff
@@ -160,5 +169,19 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
...
 	})
 
+	f.StringVar(&cli.StringVar{
+		Name:   "jwks-uri-cache-seconds",
+		Target: &cfg.JWKSURICacheSeconds,
+		EnvVar: "JWKS_URI_CACHE_SECONDS",
+		Usage:  `The number of seconds to cache the jwks_uri from an OIDC token issuer's OpenID Configuration before retrieving fresh ones. Defaults to 4 hours.`,
+	})
+
+	f.StringVar(&cli.StringVar{
```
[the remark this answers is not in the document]
Celdir: Done

pkg/server/jwk_resolver.go
Celdir: Updated the caching to directly cache `issuer -> jwks` with the configurable duration. 

Sorry for the initial complexity, I think it was mostly due to me not understanding all the tools in the `jwk` library and thinking I _needed_ to use `jwk.Cache`, which caches on `jwks_url`.

pkg/server/config.go
```diff
@@ -160,5 +161,21 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
...
+		EnvVar:  "JWKS_CACHE_DURATION",
+		Usage:   `The duration to cache the JWKS for an OIDC token issuer. Defaults to 4 hours.`,
+		Default: 4 * time.Hour,
+	})
+
+	f.StringSliceVar(&cli.StringSliceVar{
+		Name:    "issuer-allowlist",
+		Target:  &cfg.IssuerAllowlist,
+		EnvVar:  "ISSUER_ALLOWLIST",
+		Usage:   `The list of OIDC token issuers that GitHub Token Minter will accept. Format is a comma-separated list of URLs. Defaults to GitHub and Google issuers.`,
```
sethvargo, line 176:
```suggestion
		Usage:   `The list of OIDC token issuers that GitHub Token Minter will accept. Format is a comma-separated list of URLs.`,
```

And technically this accepts a comma-separated list and/or the flag can be specified multiple times.

pkg/server/jwk_resolver.go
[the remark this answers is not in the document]
sethvargo: It's not worth the optimization, especially given N is very small.

pkg/server/jwk_resolver.go
sethvargo: All good - sorry for all the back-and-forth.

sethvargo (approved):

PR #145, opened 2024-10-07
feat: support non-GitHub OIDC tokens

Celdir:
* Extract minty config GitHub repository from token's `aud` field

### DIFF
--- a/pkg/server/claims.go
+++ b/pkg/server/claims.go
@@ -20,8 +20,9 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/abcxyz/github-token-minter/pkg/config"
 	"github.com/lestrrat-go/jwx/v2/jwt"
+
+	"github.com/abcxyz/github-token-minter/pkg/config"
 )
 
 // JWTParser is an object that is responsible for parsing
--- a/pkg/server/claims.go
+++ b/pkg/server/claims.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/abcxyz/github-token-minter/pkg/config"
 	"github.com/lestrrat-go/jwx/v2/jwt"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/server/claims.go
```diff
@@ -157,6 +159,18 @@ func parsePrivateClaims(oidcToken jwt.Token) (*oidcClaims, error) {
...
 
+func extractRepository(oidcToken jwt.Token) (string, error) {
+	if oidcToken.Issuer() == config.GitHubIssuer {
+		return requiredClaim(oidcToken, "repository")
+	}
+
+	if len(oidcToken.Audience()) != 1 {
+		return "", fmt.Errorf("non-github OIDC token's audience field should have exactly one entry of a repository containing a minty config")
+	}
+	scopeRepository, _ := strings.CutPrefix(oidcToken.Audience()[0], "https://github.com/")
```
sailorlqh, line 170: I might be wrong, but could there be some possibility that the `Audience` doesn't have `https://github.com/` as prefix? Should we handle case like that?

sailorlqh (approved):

pkg/server/claims.go
Celdir: The supported formats are `https://github.com/org/repo` or simply `org/repo` and this code handles both by trimming the github URL prefix off if it is present.

sailorlqh: Thanks!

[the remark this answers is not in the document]
Celdir: So, my understanding is that GitHub issued OIDC tokens have the `repository` claim and we just read the repo from that field. But non-GitHub issued tokens (e.g. one issued by a GCP service account) do not have that claim and include very minimal metadata, but when you generate the token you can set the `aud` field to anything (at least in GCP). So we will expect the audience field to point to the repository where the relevant minty config lives, either in `org/repo` format or `https://github.com/org/repo` format. This line just normalizes the audience to `org/repo` format. 

The design doc (go/ghtm-v2-design) actually mentions using the audience field for this purpose. 
> OIDC tokens that are not sourced from GitHub will use the aud (audience) field to denote the scope of the request. The scope denotes where to find the configuration information that determines if a request for elevated privileges is valid. This will typically be something like abcxyz/infra-gcp or a full GitHub URL like https://github.com/abcxyz/infra-gcp.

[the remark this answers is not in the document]
bradegler: Yes, the aud field is the correct choice for non-github tokens. I was just wondering if we should normalized on that and ignore the `repository` field on the GitHub tokens but that's probably a more involved change than we need to do right now.

Thanks.

bradegler (approved):

PR #148, opened 2024-10-09
Integration test for gcloud token support

Celdir:
Integration test that calls github-token-minter with a Google-issued OIDC token and verifies that it receives a valid GitHub token in response

### DIFF
--- a/.github/actions/mint-token/action.yml
+++ b/.github/actions/mint-token/action.yml
@@ -17,9 +17,6 @@ inputs:
   requested_permissions:
     description: 'Permission request information in the form {"repositories":["github-token-minter"],"permissions":{"issues":"read"}}'
     required: true
-  use_gcloud_token:
-    description: 'Call github-token-minter service with OIDC token issued by GCP service account instead of GitHub'
-    default: 'false'
 
 outputs:
   token:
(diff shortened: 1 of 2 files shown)


.github/actions/mint-token/action.yml
```diff
@@ -38,17 +41,31 @@ runs:
         id_token_audience: '${{ inputs.service_audience }}'
         id_token_include_email: true
 
+    - id: 'gettoken'
```
verbanicm, line 44: From comment above, move this into the CI workflow as a separate job and then use the resulting token to call the minty endpoint manually. Use the token response to call the github api (this is strange because were in a workflow but /shrug).

Also, can you make sure to use github-script and set the response of the minty call as a secret so we dont leak the token.

.github/actions/mint-token/action.yml
```diff
@@ -17,6 +17,9 @@ inputs:
   requested_permissions:
     description: 'Permission request information in the form {"repositories":["github-token-minter"],"permissions":{"issues":"read"}}'
     required: true
+  use_gcloud_token:
```
verbanicm, line 20: This seems like the incorrect way to use this. We should always use the GitHub id_token from a GitHub action (the identity is the workflow).

Instead the token would be coming from a service running on GCP. I think your integration test should instead generate an id_token for a service account and send it to the minty endpoint. See comment below.

bradegler: I agree. We shouldn't be modifying the GitHub action at all. You can turn this test into a simple `curl` with the token from GCP in the correct header.

Celdir: Moved the test into `ci.yml` as a couple of `curl` calls with the proper tokens.

.github/actions/mint-token/action.yml
Celdir: Moved the test into the CI workflow. I'm not quite sure how to set the response as a secret using github-script, but I think the way I did things with curl wouldn't expose the token?

.github/workflows/ci.yml
```diff
@@ -153,10 +153,49 @@ jobs:
...
+          token_format: 'id_token'
+          id_token_audience: 'abcxyz/github-token-minter'
+          id_token_include_email: true
+
+      - id: 'mint-and-verify-gcloud-token'
+        run: |
+          token=$(curl --fail \
+            -X POST \
+            -H "Content-Type: application/json" \
+            -H "X-GitHub-OIDC-Token: ${{ steps.get-oidc-caller-token.outputs.id_token }}" \
```
verbanicm, line 193: As a follow up, does it make sense to still look for a header named `X-GitHub-OIDC-Token`? Maybe just rename it to X-OIDC-Token?

@bradegler thoughts?

.github/workflows/ci.yml
```diff
@@ -153,10 +153,49 @@ jobs:
...
         run: |
           curl --fail \
             -H "Accept: application/vnd.github+json" \
-            -H "Authorization: Bearer ${{ steps.mint-token.outputs.token }}"\
+            -H "Authorization: Bearer ${{ steps.mint-github-token.outputs.token }}"\
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/abcxyz/github-token-minter/issues/events
+
+      - id: 'auth'
+        name: 'Authenticate to Google Cloud'
```
verbanicm, line 165: Can you document why this is needed, if someone comes to read this it might be confusing as to why we are doing the auth twice.

First token is to generate an id token that impersonates a service account to call the minty endpoint.

Second token is to generate an id token that impersonates a service account to  token requesting access to the minty repository.

.github/workflows/ci.yml
verbanicm: I think we would need to update the action then too.

.github/workflows/ci.yml
Celdir: Added comments explaining the two separate tokens

verbanicm (approved):

pdewilde (approved):

.github/workflows/ci.yml
Celdir: Happy to do this as a followup if we want! Resolving this now so I can merge :)

bradegler: Let's leave the existing and add a new one. I don't think we have anything that calls minty outside of the action but just in case I'd rather be backwards compatible.

PR #151, opened 2024-10-15
MintyCfg - Allow Google token from WIF Service Account for integ scope

Celdir:
Allow Google token from WIF Service Account to call minty - prereq for integration test.

### DIFF
--- a/.github/minty.yaml
+++ b/.github/minty.yaml
@@ -8,7 +8,7 @@ rule:
       assertion.repository_id == '576289489'
     ) || (
       assertion.iss == issuers.google &&
-      assertion.sub == '112892967544648233217' # integration WIF service account 
+      assertion.sub == '112892967544648233217' # integration WIF service account
     )
 
 scope:
@@ -29,7 +29,7 @@ scope:
       if: |-
         assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml") || (
           assertion.iss == issuers.google &&
-          assertion.sub == "112892967544648233217" # integration WIF service account 
+          assertion.sub == "112892967544648233217" # integration WIF service account
         )
     repositories:
       - 'github-token-minter'
--- a/.github/minty.yaml
+++ b/.github/minty.yaml
@@ -8,7 +8,7 @@ rule:
       assertion.repository_id == '576289489'
     ) || (
       assertion.iss == issuers.google &&
-      assertion.sub == '112892967544648233217' # integration WIF service account
+      assertion.sub == '112892967544648233217' // integration WIF service account
     )
 
 scope:
@@ -29,7 +29,7 @@ scope:
       if: |-
         assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml") || (
           assertion.iss == issuers.google &&
-          assertion.sub == "112892967544648233217" # integration WIF service account
+          assertion.sub == "112892967544648233217" // integration WIF service account
         )
     repositories:
       - 'github-token-minter'
--- a/.github/minty.yaml
+++ b/.github/minty.yaml
@@ -4,7 +4,8 @@ rule:
   if: |-
     assertion.iss == issuers.github &&
     assertion.repository_owner_id == '93787867' &&
-    assertion.repository_id == '576289489'
+    assertion.repository_id == '576289489' ||
+    assertion.iss == issuers.google && assertion.sub == '112892967544648233217'
 
 scope:
   create-tag:
@@ -21,7 +22,7 @@ scope:
 
   integ:
     rule:
-      if: 'assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml")'
+      if: 'assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml") || assertion.iss == issuers.google && assertion.sub == "112892967544648233217"'
     repositories:
       - 'github-token-minter'
     permissions:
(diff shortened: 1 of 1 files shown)


.github/minty.yaml
```diff
@@ -4,7 +4,8 @@ rule:
   if: |-
     assertion.iss == issuers.github &&
     assertion.repository_owner_id == '93787867' &&
-    assertion.repository_id == '576289489'
+    assertion.repository_id == '576289489' ||
+    assertion.iss == issuers.google && assertion.sub == '112892967544648233217'
```
sethvargo, line 8: I think we may need to add some parens for grouping?

```cel
(
  assertion.iss == issuers.github &&
  assertion.repository_owner_id == '93787867' &&
  assertion.repository_id == '576289489'
) || (
  assertion.iss == issuers.google &&
  assertion.sub == '112892967544648233217'
)
```

sethvargo, line 8: Do we really want to allow _anything_ with that subject to mint a token? What is that integer?

Celdir: That subject is the unique ID for the integration test project's WIF service account. Unfortunately GCP-minted tokens don't seem to include a lot of metadata to work with.

Celdir: We could also check that the audience points to this repo?

verbanicm: I think this is the best we can do, according to `https://accounts.google.com/.well-known/openid-configuration` I dont see any values to use to identify the caller other than the service account number.

sethvargo: Could we at least add a comment inline

verbanicm: We could use email? Service accounts have their number built into the email right?

verbanicm: nvm, thats project number

sethvargo: Email would be better. Emails don't have ServiceAccountIds; they only have project numbers.

verbanicm: Should we worry about squatting, seems project number would mitigate that?

.github/minty.yaml
```diff
@@ -2,9 +2,14 @@ version: 'minty.abcxyz.dev/v2'
...
-    assertion.iss == issuers.github &&
-    assertion.repository_owner_id == '93787867' &&
-    assertion.repository_id == '576289489'
+    (
+      assertion.iss == issuers.github &&
+      assertion.repository_owner_id == '93787867' &&
+      assertion.repository_id == '576289489'
+    ) || (
+      assertion.iss == issuers.google &&
+      assertion.sub == '112892967544648233217' # integration WIF service account
```
verbanicm, line 11: I am not sure you can add comments inline CEL like this.....curious to see

.github/minty.yaml
Celdir: Will change this to use email, but first I need to put out a PR to make token minter parse the `email` claim from tokens. 

Service account emails have project IDs but not project numbers, e.g. `test-dmearl-sa@dmearl-gma-test-82.iam.gserviceaccount.com`. Project IDs can eventually be reused, which could be a security risk. What do we think?

.github/minty.yaml
bradegler: Yeah, that won't parse correctly:

```
[github-token-minter] [remotes/origin/dmearl/mintycfg-allow-google-token] [$] λ go run cmd/minty/main.go config validate --minty-file=.github/
minty.yaml


-- Error --
failed to compile CEL expressions in config: error compiling configuration ruleset: failed to compile ruleset: failed to compile CEL expression: ERROR: <input>:7:44: Syntax error: token recognition error at: '#'
 |   assertion.sub == '112892967544648233217' # integration WIF service account
 | ...........................................^
ERROR: <input>:7:46: Syntax error: mismatched input 'integration' expecting ')'
 |   assertion.sub == '112892967544648233217' # integration WIF service account
 | .............................................^

{"time":"2024-10-16T10:50:07.956415-04:00","severity":"ERROR","message":"process exited with error","error":"error executing validation: configuration failed to compile"}
exit status 1
```

Celdir: Fixed the inline CEL comments by using `//`. 

```
dmearl@dmearl-dev:~/code/abcxyz/github-token-minter$ go run cmd/minty/main.go config validate --minty-file=.github/minty.yaml

#################


Configuration and all rules compiled successfully
Config:
- Version: minty.abcxyz.dev/v2
- Rule.If: (
  assertion.iss == issuers.github &&
  assertion.repository_owner_id == '93787867' &&
  assertion.repository_id == '576289489'
) || (
  assertion.iss == issuers.google &&
  assertion.sub == '112892967544648233217' // integration WIF service account
)
- Scopes:
  - integ:
    - Rule.If: assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/ci.yml") || (
  assertion.iss == issuers.google &&
  assertion.sub == "112892967544648233217" // integration WIF service account
)
    - Repositories: [github-token-minter]
    - Permissions: map[issues:read]
  - create-tag:
    - Rule.If: assertion.job_workflow_ref == "abcxyz/pkg/.github/workflows/create-tag.yml@refs/heads/main" &&
assertion.workflow_ref.startsWith("abcxyz/github-token-minter/.github/workflows/create-tag.yml") &&
assertion.ref == 'refs/heads/main' &&
assertion.event_name == 'workflow_dispatch'
    - Repositories: [github-token-minter]
    - Permissions: map[contents:write]

#################

Configuration parsed and loaded successfully

```

.github/minty.yaml
Celdir: Done

Celdir: Sticking with SA number for now, but the email claim is exposed now so we can switch to that easily if we want to

verbanicm (approved):

.github/minty.yaml
sethvargo: Project IDs cannot be reused

PR #152, opened 2024-10-15
Parse email claim from OIDC token

Celdir:
Parse email claim from OIDC token, but only when `email_verified` is `true`.

### DIFF
--- a/pkg/server/claims.go
+++ b/pkg/server/claims.go
@@ -39,28 +39,33 @@ type JWTParser struct {
 // more usable forms such and org and repo names in the
 // correct format.
 type oidcClaims struct {
-	Audience          []string `json:"aud"`
-	Subject           string   `json:"sub"`
-	Issuer            string   `json:"iss"`
-	Ref               string   `json:"ref"`
-	RefType           string   `json:"ref_type"`
-	Sha               string   `json:"sha"`
-	Repository        string   `json:"repository"`
-	RepositoryID      string   `json:"repository_id"`
-	RepositoryOwner   string   `json:"repository_owner"`
-	RepositoryOwnerID string   `json:"repository_owner_id"`
-	RunID             string   `json:"run_id"`
-	RunNumber         string   `json:"run_number"`
-	Actor             string   `json:"actor"`
-	ActorID           string   `json:"actor_id"`
-	EventName         string   `json:"event_name"`
-	Workflow          string   `json:"workflow"`
-	WorkflowRef       string   `json:"workflow_ref"`
-	WorkflowSha       string   `json:"workflow_sha"`
-	JobWorkflowRef    string   `json:"job_workflow_ref"`
-	JobWorkflowSha    string   `json:"job_workflow_sha"`
-	ParsedOrgName     string   `json:"parsed_org_name"`
-	ParsedRepoName    string   `json:"parsed_repo_name"`
+	Audience []string `json:"aud"`
+	Subject  string   `json:"sub"`
+	Issuer   string   `json:"iss"`
+
+	// GitHub claims
+	Ref               string `json:"ref"`
+	RefType           string `json:"ref_type"`
+	Sha               string `json:"sha"`
+	Repository        string `json:"repository"`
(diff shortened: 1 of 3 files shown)


sethvargo (approved):

pkg/server/claims_test.go
```diff
@@ -23,26 +23,29 @@ import (
 	"github.com/abcxyz/pkg/testutil"
 )
 
-func TestTokenClaimString(t *testing.T) {
+type TestCase[T any] struct {
```
sethvargo, line 26: It's best to inline this structure instead in the test.

pkg/server/claims_test.go
```diff
@@ -80,14 +83,43 @@ func TestTokenClaimString(t *testing.T) {
 			expErrMsg: "",
 		},
 	}
+	boolCases := []TestCase[bool]{
```
sethvargo, line 86: I'm not sure why we need to break these apart. It's normal to have hundreds or thousands of table-driven tests (https://github.com/sethvargo/go-envconfig/blob/main/envconfig_test.go). If you need/want to namespace things, you can use slashes in the names.

Celdir: I only broke them apart because I didn't know how to have a slice of generic objects with different instantiated types. I didn't realize I could use `any`, thanks!

pkg/server/claims_test.go
Celdir: Done

pkg/server/claims_test.go
Celdir: Refactored the test to have all the cases defined in the same place with an inline struct. Unfortunately I still had to have a helper function which calls `tokenClaim[T]` because Go couldn't infer the type otherwise.

pkg/server/claims.go
```diff
@@ -139,22 +145,27 @@ func parsePrivateClaims(oidcToken jwt.Token) (*oidcClaims, error) {
...
+	claims.EventName = optionalClaim[string](oidcToken, "event_name")
+	claims.Workflow = optionalClaim[string](oidcToken, "workflow")
+	claims.WorkflowRef = optionalClaim[string](oidcToken, "workflow_ref")
+	claims.WorkflowSha = optionalClaim[string](oidcToken, "workflow_sha")
+	claims.JobWorkflowRef = optionalClaim[string](oidcToken, "job_workflow_ref")
+	claims.JobWorkflowSha = optionalClaim[string](oidcToken, "job_workflow_sha")
+
+	emailVerified := optionalClaim[bool](oidcToken, "email_verified")
+	if emailVerified {
+		claims.Email = optionalClaim[string](oidcToken, "email")
```
drevell, line 167: We should probably have a test for this special behavior

pkg/server/claims_test.go
```diff
@@ -87,13 +95,19 @@ func TestTokenClaimString(t *testing.T) {
...
-			}
-			if msg := testutil.DiffErrString(err, tc.expErrMsg); msg != "" {
-				t.Fatalf(msg)
-			}
+			runCase(t, tc.token, tc.value, tc.required, tc.want, tc.expErrMsg)
 		})
 	}
 }
+
+func runCase[T any](t *testing.T, token jwt.Token, value string, required bool, want T, expErrMsg string) {
```
drevell, line 103: I think this can be re-inlined now that it's only called in one place

drevell (changes requested):

pkg/server/claims_test.go
Celdir: Re-inlined by instantiating with `any` 🤮

sethvargo (approved):

pkg/server/claims.go
sethvargo: Agree

Celdir: Added test for this and also just for this function in general.

drevell: Is there a test where email_verified is false? I'm not seeing it

drevell (changes requested):

drevell (approved):

PR #155, opened 2024-10-18
Rename Auth header to X-OIDC-Token and maintain backwards compatability

Celdir:


pkg/server/token_minter.go
```diff
@@ -121,10 +122,14 @@ func (s *TokenMinterServer) processRequest(r *http.Request) *apiResponse {
 	logger := logging.FromContext(ctx)
 
 	// Retrieve the OIDC token from a header.
-	oidcHeader := r.Header.Get(AuthHeader)
+	oidcHeader := r.Header.Get(AuthHeaderV2)
+	// Check old auth header for backwards compatibility
+	if oidcHeader == "" {
+		oidcHeader = r.Header.Get(AuthHeaderV1)
+	}
```
sethvargo, line 129: Two musts and one maybe:

1. (Must) I'd like to make sure we actually remove this code in the future. The best way to do that is to add a self-breaking test after a certain date or when we bump the API version: go/chp-testing#self-breaking

1. (Must) I'd like to see tests for (a) the old header (b) the old header and the new header [uses the new header]

1. (Maybe) Should it be an error or warning if both headers are supplied?

Celdir: Took care of 1 and 2. I arbitrarily chose 12/13 as the deadline but let me know if there's a better one. 

Personally I don't think 3 is necessary since we plan on deprecating the old header anyway, I think it's fine if we just give the new one priority.

sethvargo (approved):

PR #156, opened 2024-10-21
Revert header change in action

Celdir:
Fixing minty workflow


lock14 (approved):

.github/actions/mint-token/action.yml
```diff
@@ -54,7 +54,7 @@ runs:
               body: REQUESTED_PERMISSIONS,
               headers: {
                 'Content-Type': 'application/json',
-                'X-OIDC-Token': idToken,
+                'X-GitHub-OIDC-Token': idToken,
```
sethvargo, line 57: FWIW, we could just send both headers for awhile. That will get new minty and old minty versions.

PR #159, opened 2024-10-23
feat: updated minty action to expire token in post processing

bradegler:
Moved the call to the minty server and into a new javascript action called `minty`. I then added a post section that calls the GitHub API to revoke the token in the cleanup phase.

I will follow up with changes to the `mint-token` action to call `minty` once the `minty` action gets on main.


.github/actions/minty/main.js
```diff
@@ -0,0 +1,49 @@
+const core = require('@actions/core');
```
verbanicm, line 1: We should add license, all over

.github/actions/minty/post.js
```diff
@@ -0,0 +1,18 @@
+const core = require('@actions/github');
+const github = require('@actions/github');
```
verbanicm, line 2: Same import?

.github/workflows/ci.yml
```diff
@@ -153,6 +153,15 @@ jobs:
           service_url: '${{ env.INTEGRATION_SERVICE_URL }}'
           requested_permissions: '{"scope":"integ","repositories":["github-token-minter"],"permissions":{"issues":"read"}}'
 
+      - id: 'mint-github-token-2'
```
verbanicm, line 156: Does the existing action ned to be updated? I think the inputs are different right?

.github/actions/minty/main.js
bradegler: Yes, that's part of the final cleanup. Do you feel that this implementation is correct? Mostly just looking for early input on if this is directionally the "right" way to do this.

.github/actions/minty/post.js
bradegler: `main.js` and `post.js` are loaded at separate times. They both need imports for the things they need.

.github/workflows/ci.yml
bradegler: Yes, this was just a test to make sure that it works conceptually. I'll need to land the `minty` action on main, then update the original action to look like `mint-token-2`.

.github/actions/minty/post.js
verbanicm: No I mean you are importing the same thing twice. I think you want `const core = require('@actions/core')` instead.

bradegler: oh geez, yeah, you are right

.github/actions/minty/main.js
```diff
@@ -0,0 +1,65 @@
...
+
+    // get the response as text so we can try to parse JSON
+    const responseText = await response.text();
+
+    if (response.ok) {
+      try {
+        // try to parse response as JSON for older versions of github-token-minter
+        // expecting `{ "token": "TOKEN" }` format
+
+        const resp = JSON.parse(responseText);
```
verbanicm, line 44: I would use `core.setSecret(resp.token)` prior to setting the output on this response value so its masked in logs.

.github/actions/minty/main.js
```diff
@@ -0,0 +1,65 @@
...
+        core.setOutput('token', resp.token);
+      } catch (err) {
+        // we didnt get a JSON response, response body contains the token
+        // so just use the response text
+
+        const token = responseText;
```
verbanicm, line 50: Same here.

.github/actions/minty/main.js
verbanicm: Overall process LGTM.

.github/actions/minty/main.js
```diff
@@ -0,0 +1,65 @@
...
+        core.setOutput('token', token);
+        core.exportVariable('minty-token', token);
```
verbanicm, line 52: nit: `MINTY_TOKEN`

.github/actions/minty/main.js
bradegler: 👍

.github/actions/minty/dist-main/index.js
verbanicm, line 1: nit: I think we use dist/main/index.js in our google github actions.

verbanicm (approved):

.github/actions/minty/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "main.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build-min": "ncc build -o dist/main -m main.js && ncc build -o dist/post -m post.js",
+    "build": "ncc build -o dist/main main.js && ncc build -o dist/post post.js"
+  },
+  "author": "",
+  "license": "ISC",
```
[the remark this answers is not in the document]
bradegler: fixed

PR #163, opened 2025-01-14
feat: private key import

capri-xiyue:
I planned to write private key import as a part of minty cli. The logic is not complete yet. Just add some function as the entrypoint.

### DIFF
--- a/pkg/privatekey/config.go
+++ b/pkg/privatekey/config.go
@@ -90,7 +90,7 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 	f.StringVar(&cli.StringVar{
 		Name:   "private-key",
 		Target: &cfg.PrivateKey,
-		Usage:  `The private key file to import. By default accept a filepath, and if input is exactly "-", read the value from stdin instead`,
+		Usage:  `The private key to import. By default accept a filepath, and if input is exactly "-", read the value from stdin instead`,
 	})
 
 	return set
--- a/pkg/privatekey/config.go
+++ b/pkg/privatekey/config.go
@@ -1,0 +1,98 @@
+// Copyright 2025 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package privatekey
+
+import (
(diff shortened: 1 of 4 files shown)


pkg/server/token_minter_test.go
```diff
@@ -516,7 +516,7 @@ func TestAllowRequestAllRepos(t *testing.T) {
 func TestAuthHeaderV1DeprecationDeadline(t *testing.T) {
 	t.Parallel()
 
-	keepUntil := time.Date(2024, 12, 13, 0, 0, 0, 0, time.UTC)
+	keepUntil := time.Date(2025, 12, 13, 0, 0, 0, 0, time.UTC)
```
capri-xiyue, line 519: I'm not sure whether it's safe to do this now. I guess this UT is to remind devs in github token minter to do some change?

pkg/privatekey/config.go
```diff
@@ -0,0 +1,98 @@
...
+		Name:   "key",
+		Target: &cfg.Key,
+		EnvVar: "KEY",
+		Usage:  `the name of the key.`,
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:   "private-key-file",
+		Target: &cfg.PrivateKeyFile,
+		EnvVar: "PRIVATE_KEY_FILE",
```
sethvargo, line 93: The problem with requiring a filepath is that it requires any automation to write the key to disk before running the command. I think we need the ability to pass the key directly.

pkg/server/token_minter_test.go
sethvargo: This was the deprecation timeline for the old field. We need to go delete the old code :)

/cc @shankiyani I think you did the original deprecation work on this.

pkg/privatekey/config.go
capri-xiyue: Probably there is some gap about the CUJ. 
The CUJ we plan is that 
1. users download the github private key to local disk
2. users use the minty cli command to import the private the key to cloud kms

We don't plan the automation to write the key to disk. Users just click github UI to download the private key. The ability to pass the key directly is blocked now as there is no official github api to do it.

sethvargo: I think we can support both. The patterns I usually use are:

- Make the variable accept a string, and if the string starts with `@`, assume it's a filepath (and read contents from disk)

- Make the variable accept a filepath, and if input is exactly `-`, read the value from stdin instead

capri-xiyue: Got it. I will use the approach to make the variable accept a filepath, and if input is exactly -, read the value from stdin instead as I think filepath is more common here as when users download the github private key, by default it is a file.

pkg/server/token_minter_test.go
shankiyani: I recall us having this conversation but I believe @Celdir ended up taking this work up when I left last year.

pkg/privatekey/config.go
capri-xiyue: I've changed the name to indicate it support both. Will implement it in later PR.

sailorlqh (approved):

PR #168, opened 2025-01-16
feat: created key function added

capri-xiyue:
part of https://github.com/abcxyz/github-token-minter/issues/170

This PR added the create cloud kms key function.

### DIFF
--- a/pkg/privatekey/kms.go
+++ b/pkg/privatekey/kms.go
@@ -33,8 +33,8 @@ func NewKeyServer(ctx context.Context, kms *kms.KeyManagementClient) (*KeyServer
 	return &KeyServer{kms}, nil
 }
 
-// CreateKeyRingIfNotExists get the existing key ring or create a key ring if it doesn't exist.
-func (s *KeyServer) CreateKeyRingIfNotExists(ctx context.Context, projectID, location, keyRing string) (*kmspb.KeyRing, error) {
+// GetOrCreateKeyRing get the existing key ring or create a key ring if it doesn't exist.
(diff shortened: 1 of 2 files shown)


pkg/privatekey/kms.go
```diff
@@ -61,3 +61,39 @@ func (s *KeyServer) CreateKeyRingIfNotExists(ctx context.Context, projectID, loc
...
+				},
+				SkipInitialVersionCreation: true,
+			}
+
+			createdKey, err := s.kms.CreateCryptoKey(ctx, req)
+			if err != nil {
+				return nil, fmt.Errorf("failed to create key with key path %q: %w", keyPath, err)
+			}
+			return createdKey, nil
+		} else {
```
drevell, line 93: Unnecessary "else" here

pkg/privatekey/kms.go
```diff
@@ -61,3 +61,39 @@ func (s *KeyServer) CreateKeyRingIfNotExists(ctx context.Context, projectID, loc
 	// Key ring already exists.
 	return fetchedKeyRing, nil
 }
+
+// CreateKeyIfNotExists create a key if it doesn't exist.
```
drevell, line 65: The name and the comment do not explain what happens if the key does exist. Maybe name it GetOrCreateKey()?

capri-xiyue:
> Could you add a description to this PR? I'm not sure what we're intending to accomplish and how it ties into a feature/bug

updated it.

pkg/privatekey/kms.go
capri-xiyue: updated.

pkg/privatekey/kms.go
capri-xiyue: fixed.

Issue #170, opened 2025-01-16
Add private key import feature in github-token-minter cli


capri-xiyue:
### TL;DR

Add private key import feature in github-token-minter cli to facilitate github token minter admins to import the github private key into cloud kms and then later github-token-minter can use the imported private key to sign the token.

### Detailed design

```markdown

```

### Alternatives considered

```markdown

```

PR #176, opened 2025-01-22
feat: add import key version

capri-xiyue:
part of #170 
Added import key version function

### DIFF
--- a/pkg/cli/private_key_import_cfg.go
+++ b/pkg/cli/private_key_import_cfg.go
@@ -116,7 +116,7 @@ func (c *PrivateKeyImportCommand) Run(ctx context.Context, args []string) error 
 		}
 		return nil
 	}); err != nil {
-		return fmt.Errorf("failed to check key version state to be enabled: %w", err)
+		return fmt.Errorf("failed to check import job state to be active: %w", err)
 	}
 	createdKeyVersion, err := keyServer.ImportManuallyWrappedKey(ctx, gotImportJob.GetName(), gotKey.GetName(), c.cfg.PrivateKey)
 	if err != nil {
--- a/pkg/cli/private_key_import_cfg.go
+++ b/pkg/cli/private_key_import_cfg.go
@@ -105,7 +105,7 @@ func (c *PrivateKeyImportCommand) Run(ctx context.Context, args []string) error 
 	if err != nil {
 		return fmt.Errorf("encountered error when checking state of import job: %w", err)
 	}
-	if importedJob.State != kmspb.ImportJob_ACTIVE {
+	if importedJob.GetState() != kmspb.ImportJob_ACTIVE {
 		return fmt.Errorf("import job is not in active stage")
 	}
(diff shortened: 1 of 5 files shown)


pkg/cli/private_key_import_cfg.go
```diff
@@ -89,18 +91,40 @@ func (c *PrivateKeyImportCommand) Run(ctx context.Context, args []string) error
...
+	}
+	if importedJob.GetState() != kmspb.ImportJob_ACTIVE {
+		return fmt.Errorf("import job is not in active stage, you may re-run it to retry")
+	}
+	createdKeyVersion, err := keyServer.ImportManuallyWrappedKey(ctx, gotImportJob.GetName(), gotKey.GetName(), c.cfg.PrivateKey)
+	if err != nil {
+		return fmt.Errorf("failed to import key version: %w", err)
+	}
+	logger.DebugContext(ctx, "Got key version imported", "key_version", createdKeyVersion.GetName())
+	//sleep 3 second to wait for the key version to becoming enabled.
```
sethvargo: I think we want to do this with go-retry and wait for the key to get into the state we want instead of an arbitrary sleep time.

capri-xiyue: fixed with `go-retry`

pkg/cli/private_key_import_cfg.go
```diff
@@ -89,18 +94,51 @@ func (c *PrivateKeyImportCommand) Run(ctx context.Context, args []string) error
...
+	logger.DebugContext(ctx, "Got key successfully", "key_ring", gotKey.GetName())
 	gotImportJob, err := keyServer.GetOrCreateImportJob(ctx, c.cfg.ProjectID, c.cfg.Location, c.cfg.KeyRing, c.cfg.ImportJobPrefix)
 	if err != nil {
 		return fmt.Errorf("encountered error when creating/getting import job: %w", err)
 	}
-	logger.InfoContext(ctx, "Got import job successfully\n", "import_job", gotImportJob.GetName())
-	// TODO import key version
+	logger.DebugContext(ctx, "Got import job successfully", "import_job", gotImportJob.GetName())
+
+	if err := retry.Do(ctx, backoff, func(ctx context.Context) error {
```
sethvargo, line 109: Unfortunately you can't re-use the backoff like this. The backoff is stateful, so the retry attempts from the first will get counted in the second attempt, which isn't what you want. You need to create a new backoff for each invocation of `retry.Do`. You can make it helper to inline it:

```go
func newBackoff() retry.Backoff {
  return retry.WithMaxRetries(5, retry.NewFibonacci(500*time.Millisecond))
}

// ...

retry.Do(ctx, newBackoff(), func(ctx context.Context) error {
  // ...
})
```

capri-xiyue: Thanks for catching it! Didn't know `go-retry` was stateful.

sailorlqh (approved):

PR #179, opened 2025-01-24
fix: some tokens won't parse cleanly when using the validation cli if they have arrays

bradegler:


verbanicm (approved):

pkg/mintycfg/runner.go
```diff
@@ -83,7 +83,7 @@ func Run(ctx context.Context, cfg *Config) error {
 	// if a scope and a token were provided, run evaluation against them to
 	// determine if there is a scope match and then output the scope contents
 	if cfg.Scope != "" && cfg.Token != "" {
-		var token map[string]string
+		var token map[string]interface{}
```
verbanicm, line 86: @bradegler  GAH or should this be any?

bradegler: Fixed in #180

PR #192, opened 2025-02-26
feat: Update WIF pool assertions to allow for enterprise id and/or organization id

bradegler:
This will allow us to deploy instances of minty for multiple organizations.

### DIFF
--- a/terraform/github-wif.tf
+++ b/terraform/github-wif.tf
@@ -18,8 +18,11 @@ locals {
     "attribute.actor" : "assertion.actor"
     "attribute.aud" : "assertion.aud"
     "attribute.repository_owner_id" : "assertion.repository_owner_id"
+    "attribute.enterprise_id" : "assertion.enterprise_id"
   }
-  wif_attribute_condition = var.github_owner_id != "" ? "attribute.repository_owner_id == \"${var.github_owner_id}\"" : ""
+  wif_attr_org_id        = var.github_owner_id != "" ? "attribute.repository_owner_id == \"${var.github_owner_id}\"" : ""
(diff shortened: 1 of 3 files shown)


verbanicm (commented):
Is the is useful? Why not just make it public then?

bradegler:
> Is the is useful? Why not just make it public then?

Sorry, I'm not following what you are asking.

sethvargo:
I'm really not in favor or omitting the owner id for federation. What is the use case?

bradegler:
I need to run Minty for 300 GitHub orgs. I can't do that if the wif pool blocks 299 of them and I don't want to run 300 instances.

bradegler:
@sethvargo - I added in the `enterprise_id` as an option. Please review when you have a min.

terraform/github-wif.tf
```diff
@@ -43,7 +46,7 @@ resource "google_iam_workload_identity_pool_provider" "default" {
   description                        = "${var.name} OIDC identity provider"
 
   attribute_mapping   = local.wif_attribute_mapping
-  attribute_condition = local.wif_attribute_condition
+  attribute_condition = var.github_owner_id != "" && var.github_enterprise_id != "" ? local.wif_attr_org_id + " && " + local.wif_attr_enterprise_id : local.wif_attr_org_id + local.wif_attr_enterprise_id
```
verbanicm, line 49: I find this a bit confusing, I would add a comment for posterity.

bradegler: I added a comment explaining how that works.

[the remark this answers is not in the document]
bradegler: Yeah, that's simpler - thanks!

PR #194, opened 2025-03-06
feat: extracted GitHub specific code into an isolated class

bradegler:
* Moved the GitHub specific code behind an interface to make room for additional implementations (GitLab).
* Started a means to specify multiple GitHub Apps to be used. 
 
The multi App work is not completed as there are two things left to solve in a follow up PR that I wouldn't mind some feedback on. The current implementation just loops through them which is dumb.

* First, how to deal with configuration for multiple Apps. I'm not sure if I want multiple command line options as they would need to be pairs of APP_ID/PRIVATE_KEY so possibly some kind of config file is warranted? This needs to be thought through still.
* Second, how to determine which app to pick for any given request. I'm thinking this is probably tied to the request object but I'm still trying to figure out how to do this correctly.

### DIFF
--- a/pkg/config/config_loader_test.go
+++ b/pkg/config/config_loader_test.go
@@ -16,6 +16,8 @@ package config
 
 import (
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -25,6 +27,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-github/v64/github"
 
+	"github.com/abcxyz/github-token-minter/pkg/server/source"
 	"github.com/abcxyz/pkg/testutil"
 )
 
@@ -58,7 +61,7 @@ func TestGitHubInRepoConfigFileLoader(t *testing.T) {
 				w.WriteHeader(200)
 				fmt.Fprint(w, string(raw))
 			},
-			want:      &Config{},
+			want:      nil,
 			expErr:    false,
 			expErrMsg: "",
 		},
@@ -175,21 +178,39 @@ rule:
 	ctx := t.Context()
 	for _, tc := range cases {
 		mux := http.NewServeMux()
+		mux.Handle("GET /repos/test_org/test_repo/installation", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			fmt.Fprintf(w, `{"access_tokens_url": "http://%s/app/installations/123/access_tokens"}`, r.Host)
+		}))
+		mux.Handle("POST /app/installations/123/access_tokens", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(201)
+			fmt.Fprintf(w, `{"token": "this-is-the-token-from-github"}`)
+		}))
 		mux.Handle("/api/v3/repos/test_org/test_repo/contents/minty.yaml", tc.handler)
-		srv := httptest.NewServer(mux)
-		t.Cleanup(srv.Close)
+		fakeGitHub := httptest.NewServer(mux)
+		t.Cleanup(fakeGitHub.Close)
 
-		client, _ := github.NewClient(nil).WithEnterpriseURLs(srv.URL, srv.URL)
+		rsaPrivateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		ghAppCfg := source.GitHubAppConfig{
+			AppID:  "app-id",
+			Signer: rsaPrivateKey,
+		}
+
+		sourceSystem, err := source.NewGitHubSourceSystem(ctx, []*source.GitHubAppConfig{&ghAppCfg}, fakeGitHub.URL)
+		if err != nil {
+			t.Fatal(err)
+		}
 
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
-			loader := ghInRepoConfigFileLoader{
-				configPath: tc.path,
-				ref:        tc.ref,
-				provider: func(ctx context.Context, org, repo string) (*github.Client, error) {
-					return client, nil
-				},
+			loader := &inRepoConfigFileLoader{
+				configPath:   tc.path,
+				ref:          tc.ref,
+				sourceSystem: sourceSystem,
 			}
 
 			got, err := loader.Load(ctx, tc.org, tc.repo)
@@ -319,17 +340,23 @@ func TestConfigFileLoaderSource(t *testing.T) {
 		},
 		{
 			name:   "in repo config loader",
-			loader: &ghInRepoConfigFileLoader{configPath: ".test/minty.yaml"},
+			loader: &inRepoConfigFileLoader{configPath: ".test/minty.yaml", sourceSystem: &testSourceSystem{}},
 			org:    "test_org",
 			repo:   "test_repo",
-			want:   "https://github.com/test_org/test_repo/.test/minty.yaml",
+			want:   "https://test.com/test_org/test_repo/.test/minty.yaml",
 		},
 		{
-			name:   "fixed repo config loader",
-			loader: &fixedRepoConfigFileLoader{repo: "not_test_repo", loader: &ghInRepoConfigFileLoader{configPath: "minty.yaml"}},
-			org:    "test_org",
-			repo:   "test_repo",
-			want:   "https://github.com/test_org/not_test_repo/minty.yaml",
+			name: "fixed repo config loader",
+			loader: &fixedRepoConfigFileLoader{
+				repo: "not_test_repo",
+				loader: &inRepoConfigFileLoader{
+					configPath:   "minty.yaml",
+					sourceSystem: &testSourceSystem{},
+				},
+			},
+			org:  "test_org",
+			repo: "test_repo",
+			want: "https://test.com/test_org/not_test_repo/minty.yaml",
 		},
 	}
 
@@ -343,4 +370,18 @@ func TestConfigFileLoaderSource(t *testing.T) {
 			}
 		})
 	}
+}
+
+type testSourceSystem struct{}
+
+func (s *testSourceSystem) MintAccessToken(ctx context.Context, org, repo string, repositories []string, permissions map[string]string) (string, error) {
+	return "token", nil
+}
+
+func (s *testSourceSystem) RetrieveFileContents(ctx context.Context, org, repo, filePath, ref string) ([]byte, error) {
+	return []byte{}, nil
+}
+
+func (s *testSourceSystem) BaseURL() string {
+	return "https://test.com"
 }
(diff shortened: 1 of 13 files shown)


pkg/config/config_loader_test.go
```diff
@@ -176,20 +178,31 @@ rule:
...
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
-			loader := ghInRepoConfigFileLoader{
-				configPath: tc.path,
-				ref:        tc.ref,
-				provider: func(ctx context.Context, org, repo string) (*github.Client, error) {
-					return client, nil
-				},
+			loader := inRepoConfigFileLoader{
```
sethvargo, line 202:
```suggestion
			loader := &inRepoConfigFileLoader{
```

?

pkg/server/source/github.go
```diff
@@ -0,0 +1,152 @@
...
+	client, err := kms.NewKeyManagementClient(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to setup client: %w", err)
+	}
+	signer, err := gcpkms.NewSigner(ctx, client, keyID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create signer: %w", err)
+	}
+	return signer, nil
+}
```
sethvargo, line 52: I'm still not convinced we want this directly in here; we should really only care about accepting a `crypto.Signer` interface.

pkg/server/source/source_system.go
```diff
@@ -0,0 +1,28 @@
...
+type System interface {
+	// MintAccessToken generates a new access token on behalf of the org/rep for the given repositories with
+	// the specified permissions.
+	MintAccessToken(ctx context.Context, org, repo string, repositories []string, permissions map[string]string) (string, error)
+	// RetrieveFileContents gets the contents of the file at filePath with the specified ref
+	// from the org/repo.
+	RetrieveFileContents(ctx context.Context, org, repo, filePath, ref string) ([]byte, error)
+	// BaseURL gets the base url for the system.
+	BaseURL() string
+}
```
sethvargo, line 28:
```suggestion
type System interface {
	// MintAccessToken generates a new access token on behalf of the org/rep for the given repositories with
	// the specified permissions.
	MintAccessToken(ctx context.Context, org, repo string, repositories []string, permissions map[string]string) (string, error)
	
	// RetrieveFileContents gets the contents of the file at filePath with the specified ref
	// from the org/repo.
	RetrieveFileContents(ctx context.Context, org, repo, filePath, ref string) ([]byte, error)
	
	// BaseURL gets the base url for the system.
	BaseURL() string
}
```

pkg/server/token_minter_test.go
```diff
@@ -225,16 +225,22 @@ func TestTokenMintServer_ProcessRequest(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			githubApp, err := githubauth.NewApp("app-id", rsaPrivateKey, githubauth.WithBaseURL(fakeGitHub.URL))
+			ghAppCfg := source.GitHubAppConfig{
```
sethvargo, line 228:
```suggestion
			ghAppCfg := &source.GitHubAppConfig{
```

?

sethvargo (commented):
> 1. How to deal with configuration for multiple Apps. I'm not sure if I want multiple command line options as they would need to be pairs of APP_ID/PRIVATE_KEY so possibly some kind of config file is warranted? This needs to be thought through still.

I like inventing a protocol string for these things like `gh://<app-id>?private-key=<blob>&private-key-path=<path>`

> 2. How to determine which app to pick for any given request. I'm thinking this is probably tied to the request object but I'm still trying to figure out how to do this correctly.

I _think_ GitHub will include the _installation_ ID in the request, which you should be able to map to the app IIRC.

pkg/server/source/github.go
bradegler: I moved it up a level to the runner.

bradegler:
> > 1. How to deal with configuration for multiple Apps. I'm not sure if I want multiple command line options as they would need to be pairs of APP_ID/PRIVATE_KEY so possibly some kind of config file is warranted? This needs to be thought through still.
> 
> I like inventing a protocol string for these things like `gh://<app-id>?private-key=<blob>&private-key-path=<path>`
> 

So, if I'm following, with that scheme I'd make something like GITHUB_APPS=gh://<app id 1>?pk=blah,gh://<app id 2>?kms=blah

> > 2. How to determine which app to pick for any given request. I'm thinking this is probably tied to the request object but I'm still trying to figure out how to do this correctly.
> 
> I _think_ GitHub will include the _installation_ ID in the request, which you should be able to map to the app IIRC.

I don't think that will work. The request is coming from an action workflow which doesn't have installation information. We look up the installation from the org and repo. This code loops looking for the first install but that won't actually work. We'd likely have multiples of the apps installed.

bradegler:
I've changed the scope of this PR slightly.

I've added the ability to specify different key types on the command line, provided support for KMS and moved some of the code around to make this easier to maintain.

I limited the number of GitHub Apps to 1 for the time being until we have some time to think about how to make multiple apps work as that is not critical right now.

@sethvargo @verbanicm - please take another look

pkg/server/config.go
```diff
@@ -42,14 +45,12 @@ type Config struct {
 	IssuerAllowlist   []string
 }
 
+const (
+	SourceSystemAuthConfigRegex = `gha\:\/\/(\d+)\?(private_key|kms_id)=([\s\S]*)`
```
sethvargo, line 49: I'm not sure we want to use a regex here versus `url.Parse` and extracting what we care about. `url.Parse("gha://app-id?private_key=babdf")` will parse and you can check the Protocol and iterate over the URL Params nicely.

pkg/server/runner.go
```diff
@@ -89,3 +92,64 @@ func Run(ctx context.Context, cfg *Config) error {
...
+			}
+		} else {
+			return nil, fmt.Errorf("invalid key type: %s", keyType)
+		}
+
+		ghAppCfg := source.GitHubAppConfig{
+			AppID:  appID,
+			Signer: signer,
+		}
+		appConfigs = append(appConfigs, &ghAppCfg)
```
sethvargo, line 140:
```suggestion
		appConfigs = append(appConfigs, &source.GitHubAppConfig{
			AppID:  appID,
			Signer: signer,
		})
```

pkg/server/source/github.go
```diff
@@ -0,0 +1,130 @@
...
+	AppID  string
+	Signer crypto.Signer
+}
+
+type gitHubSourceSystem struct {
+	apps    []*githubauth.App
+	baseURL string
+}
+
+func NewGitHubSourceSystem(ctx context.Context, configs []*GitHubAppConfig, systemURL string) (System, error) {
```
sethvargo, line 39: Please doc these

pkg/server/source/source_system.go
```diff
@@ -0,0 +1,30 @@
...
+	// MintAccessToken generates a new access token on behalf of the org/rep for the given repositories with
+	// the specified permissions.
+	MintAccessToken(ctx context.Context, org, repo string, repositories []string, permissions map[string]string) (string, error)
+
+	// RetrieveFileContents gets the contents of the file at filePath with the specified ref
+	// from the org/repo.
+	RetrieveFileContents(ctx context.Context, org, repo, filePath, ref string) ([]byte, error)
+
+	// BaseURL gets the base url for the system.
+	BaseURL() string
```
sethvargo, line 29: We might want to separate API URL, HTML URL, and Base URL?

sethvargo (approved):
I'm still not in love with the tight coupling to Google Cloud KMS, but I'll defer

pkg/server/source/source_system.go
bradegler: I'll get this in a follow up

pkg/server/source/github.go
bradegler: done

pkg/server/config.go
bradegler: `url.Parse` doesn't love the fact that that private_key parameter is an actual PEM encoded PK. I tried multiple ways of escaping that string and it throws errors like `net/url: invalid control character in URL` no matter what I do. 

I've got a few more things to try but this doesn't look like it will work.

bradegler:
> I'm still not in love with the tight coupling to Google Cloud KMS, but I'll defer

Can you articulate your concerns with using KMS as our signer? Is it a rate limit question or do you think it'll have too much latency or do you dislike a dependent service that could be down or is it something else I'm not thinking of?

I really want to get the private key out of the process if I can but if you feel KMS is not the right way to do that I'd like to know more about why.

pkg/server/config.go
sethvargo: right, you'd have to hex/b64 encode the string for it to be urlsafe (and use urlsafe b64).

sethvargo:
> Can you articulate your concerns with using KMS as our signer? Is it a rate limit question or do you think it'll have too much latency or do you dislike a dependent service that could be down or is it something else I'm not thinking of?

It's not about using KMS, it's about coupling this repo to GCP KMS (versus it being a generic "signer"). If we ever had to run this on Kubernetes or AWS, we'd have other signing implementations, and it feels bloated to have all those SDKs in this particular package.

bradegler:
> > Can you articulate your concerns with using KMS as our signer? Is it a rate limit question or do you think it'll have too much latency or do you dislike a dependent service that could be down or is it something else I'm not thinking of?
> 
> It's not about using KMS, it's about coupling this repo to GCP KMS (versus it being a generic "signer"). If we ever had to run this on Kubernetes or AWS, we'd have other signing implementations, and it feels bloated to have all those SDKs in this particular package.

I must be missing something. 99% of this code uses the Signer interface. It's just the server startup code that initializes a KMS based signer. There is no other way for me to pass a Signer out of band from server startup unless you are suggesting some kind of magic that I'm not aware of.

I guess I could make raw API calls to GCP endpoints instead of using the client lib but that doesn't feel like what you are suggesting. 

I'm feeling confused on this one.

sethvargo:
> I must be missing something. 99% of this code uses the Signer interface. It's just the server startup code that initializes a KMS based signer. There is no other way for me to pass a Signer out of band from server startup unless you are suggesting some kind of magic that I'm not aware of.
> 
> I guess I could make raw API calls to GCP endpoints instead of using the client lib but that doesn't feel like what you are suggesting.
> 
> I'm feeling confused on this one.

It's fine. Let's just merge this as-is, and we can revisit supporting multiple clouds via a plugin model in the future if needed. Sorry.

bradegler:
> > I must be missing something. 99% of this code uses the Signer interface. It's just the server startup code that initializes a KMS based signer. There is no other way for me to pass a Signer out of band from server startup unless you are suggesting some kind of magic that I'm not aware of.
> > I guess I could make raw API calls to GCP endpoints instead of using the client lib but that doesn't feel like what you are suggesting.
> > I'm feeling confused on this one.
> 
> It's fine. Let's just merge this as-is, and we can revisit supporting multiple clouds via a plugin model in the future if needed. Sorry.

Thanks for the explanation. I never think of Go as a plugin kind of language so I wasn't wrapping my head around what you were trying to suggest. I appreciate you, as usual, reminding me to think differently.

PR #203, opened 2025-04-24
feat: add logic to return the scope's permissions if none are requested

bradegler:
This was a missed 2.0 feature.

A caller should be able to just request the scope and receive all of the permissions that are defined for that scope. This simplifies what a workflow writer needs to know.


pkg/server/token_minter.go
```diff
@@ -168,6 +168,11 @@ func (s *TokenMinterServer) processRequest(r *http.Request) *apiResponse {
 		return &apiResponse{http.StatusForbidden, fmt.Sprintf("no permissions available for scope %q in repository %q", request.Scope, claims.Repository), err}
 	}
 
+	// If there are no permissions in the request, use the set defined for the scope.
```
verbanicm, line 171: So if i request nothing i get...?

Celdir (approved):

pkg/server/token_minter.go
bradegler: What is defined for that scope.

The goal being that the scopes defines the configuration of what you are allowed to do with it. 

You can reduce that by requesting less if you want but in most cases you will have the workflow match what's in config. This eliminates the need for that duplication.

pkg/server/token_minter_test.go
```diff
@@ -41,6 +42,25 @@ import (
...
 
+func handleAccessTokenRequest(w http.ResponseWriter, r *http.Request) {
+	// Parse the request information
+	defer r.Body.Close()
+
+	var request tokenRequest
+	dec := json.NewDecoder(io.LimitReader(r.Body, 4_194_304)) // 4 MiB
+	if err := dec.Decode(&request); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintf(w, "error parsing request information - invalid JSON: %s", err)
```
sethvargo, line 53: Missing return

bradegler: Fixed.

sethvargo (approved):

pkg/server/token_minter.go
bradegler: We log the request object (with or without permissions) 20 lines above this code. I don't think another log message that says "no requested permissions" or whatever will provide any value since I would probably check the token request first anyhow.

PR #215, opened 2025-08-07
Rely on IssuerAllowlist when attempting to extract repository claim

shankiyani:


pkg/server/claims.go
```diff
@@ -177,31 +175,39 @@ func parsePrivateClaims(oidcToken jwt.Token) (*oidcClaims, error) {
...
-func extractRepository(oidcToken jwt.Token) (string, error) {
-	if oidcToken.Issuer() == config.GitHubIssuer {
-		return requiredClaim[string](oidcToken, "repository")
+func extractRepository(oidcToken jwt.Token, issuerAllowlist []string) (string, error) {
+	// Iterating through the allowlist to check if it matches the token's issuer.
+	issuerAllowed := false
+	for _, allowedIssuer := range issuerAllowlist {
+		if oidcToken.Issuer() == allowedIssuer {
+			issuerAllowed = true
+			break
```
verbanicm, line 195: Why not just return here instead of the intermediate variable?

pkg/server/claims.go
```diff
@@ -177,31 +175,39 @@ func parsePrivateClaims(oidcToken jwt.Token) (*oidcClaims, error) {
...
 	// the repository_owner claim is optional.
 	repoParts := strings.Split(claims.Repository, "/")
-	if len(repoParts) != 2 {
-		return nil, fmt.Errorf("'repository' claim formatted incorrectly, requires <org_name>/<repo_name> format - received [%s]", claims.Repository)
+	if len(repoParts) == 2 {
+		claims.ParsedOrgName = repoParts[0]
+		claims.ParsedRepoName = repoParts[1]
+	} else {
+		claims.ParsedOrgName = ""
+		claims.ParsedRepoName = ""
```
verbanicm, line 183: Could prefer to just set these to empty string, then the if statement just overrides it if its len is correct

shankiyani:
> What problem are you trying to solve here? Why can't you use https://pkg.go.dev/github.com/lestrrat-go/jwx/v3/jwt#WithIssuer?

Discussed offline this change is obsolete in favor of https://github.com/abcxyz/github-token-minter/pull/217

PR #230, opened 2025-09-25
fix: invalid format when requesting all permissions

bradegler:
This PR fixes an issue where a nil permissions map was sent to the GitHub API when requesting a token, causing a 500 error from GitHub

The fix initializes the permissions map to an empty map and only populates it if permissions are provided. This ensures that the request sent to GitHub is always in a valid format.

Additionally, this PR adds logging for the token request to improve observability.

### DIFF
--- a/pkg/server/source/github.go
+++ b/pkg/server/source/github.go
@@ -25,6 +25,7 @@ import (
 	"github.com/google/go-github/v64/github"
 
 	"github.com/abcxyz/pkg/githubauth"
+	"github.com/abcxyz/pkg/logging"
 )
 
 // GitHubAppConfig is a struct that contains the identifier for a GitHub App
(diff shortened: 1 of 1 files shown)


pkg/server/source/github.go
```diff
@@ -95,8 +102,12 @@ func (g *gitHubSourceSystem) MintAccessToken(ctx context.Context, org, repo stri
 	}
 	tokenRequest := githubauth.TokenRequest{
 		Repositories: repositories,
-		Permissions:  permissions,
+		Permissions:  map[string]string{},
```
verbanicm, line 105: check my understanding, if this is empty, then zero permissions are given, but if the permissions attribute was missing altogether, then all permissions are given correct?

verbanicm (approved):

pkg/server/source/github.go
bradegler: If the map is empty then we get all permissions defined on the app. To get to this point we have multiple checks to verify that it is allowed to request that.

If the `Permissions` attribute is null then GitHub throws a 500

PR #247, opened 2026-04-06
feat: add 422 retries and optimize config caching

RivocPX:
Addresses minty failures caused by caching empty configs and transient GitHub 422 errors.

- Stopped caching nil results in `cachingConfigFileLoader` to prevent 15-minute locks.
- Added a configurable Retry422 flag (default false) to retry HTTP 422 errors in RetryRoundTripper.
- Passed the retry-enabled HTTP client to the githubauth library for token requests.
- Added unit tests. Tested by `go test ./...`

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -51,6 +51,7 @@ type Config struct {
 	GitHubRequestMaxBackoff     time.Duration
 	GitHubRequestMultiplier     float64
 	GitHubRequestRetry404       bool
+	GitHubRequestRetry422       bool
 	EnforceReadOnly             bool
 }
 
@@ -82,6 +83,7 @@ func (cfg Config) LogValue() slog.Value {
 		slog.Duration("github_request_max_backoff", cfg.GitHubRequestMaxBackoff),
 		slog.Float64("github_request_multiplier", cfg.GitHubRequestMultiplier),
 		slog.Bool("github_request_retry_404", cfg.GitHubRequestRetry404),
+		slog.Bool("github_request_retry_422", cfg.GitHubRequestRetry422),
 		slog.Bool("enforce_read_only", cfg.EnforceReadOnly),
 	)
 }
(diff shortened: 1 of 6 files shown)


kevya-google (approved):
Thank you for digging in and fixing this!

pkg/server/config.go
```diff
@@ -280,6 +282,14 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 		Usage:   `Whether to retry GitHub API requests that return a 404 Not Found status.`,
 	})
 
+	f.BoolVar(&cli.BoolVar{
+		Name:    "github-request-retry-422",
+		Target:  &cfg.GitHubRequestRetry422,
+		EnvVar:  "GITHUB_REQUEST_RETRY_422",
+		Default: false,
```
bradegler: We should default this and the 404 config to `true`. It has baked long enough to know its not going to break anything.

pkg/server/source/github.go
```diff
@@ -243,6 +253,14 @@ func (r *RetryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
 			return retry.RetryableError(fmt.Errorf("server error: %d", resp.StatusCode))
 		}
 
+		if resp.StatusCode == http.StatusUnprocessableEntity && r.RetryConfig.Retry422 {
```
bradegler: nit; you could probably merge this if with the one below to prevent duplicating the body of it.

pkg/server/source/github.go
```diff
@@ -63,6 +64,15 @@ func NewGitHubSourceSystem(ctx context.Context, configs []*GitHubAppConfig, syst
 	if systemURL != "" {
 		options = append(options, githubauth.WithBaseURL(systemURL))
 	}
+	if retryConfig != nil {
```
bradegler, line 69: If you are going to move this up to when the app is initialized then you should remove it from https://github.com/abcxyz/github-token-minter/blob/main/pkg/server/source/github.go#L154C1-L166C61

pkg/server/config.go
RivocPX: Done. Both 404 and 422 are default to be true.

pkg/server/source/github.go
RivocPX: Done. Merge this with 404 below.

pkg/server/source/github.go
RivocPX: Done. Move this from `RetrieveFileContents` to `NewGitHubSourceSystem`.

bradegler (approved):

PR #251, opened 2026-04-13
feat: add server side policy enforcement

bradegler:


pkg/server/config.go
```diff
@@ -84,7 +85,6 @@ func (cfg Config) LogValue() slog.Value {
 		slog.Float64("github_request_multiplier", cfg.GitHubRequestMultiplier),
 		slog.Bool("github_request_retry_404", cfg.GitHubRequestRetry404),
 		slog.Bool("github_request_retry_422", cfg.GitHubRequestRetry422),
```
jalseth, line 87: This seems like a breaking change. Do we know if anyone  is using this?

policy/fail_safe.rego
```diff
@@ -0,0 +1,28 @@
+package minty.policy
+
+# Fail-safe: Only allow token requests from a specific enterprise, org, and repo ID.
+# Replace these values with your actual allowed IDs.
+allowed_enterprise_id := "YOUR_ENTERPRISE_ID"
+allowed_org_id := "YOUR_ORG_ID"
+allowed_repo_id := "YOUR_REPO_ID"
+
+# Deny if enterprise ID is present and does not match
+deny contains msg if {
```
jalseth, line 10: These checks will by bypassed if the token doesn't have these fields, because accessing the missing key will returned undefined and Rego eval will stop. This should not happen, but it's best practice to guard against this.

policy/read_only.rego
```diff
@@ -0,0 +1,9 @@
+package minty.policy
+
+# Ensure no configuration allows "write" permissions in this deployment.
+deny contains msg if {
```
jalseth, line 4: IIRC (not sure), configs without explicit permissions are treated as requesting all permissions that the app has. If that is true, this policy would be bypassed by having an empty permissions block.

pkg/server/config.go
bradegler: We haven't deployed a read only instance yet.

policy/fail_safe.rego
bradegler: Interesting. I'll look into how to protect against that.

policy/read_only.rego
bradegler: That is correct. I'll build a rule that handles that too, good catch.

policy/fail_safe.rego
bradegler: I changed the example policies to account for these. lmk what you think

pkg/mintycfg/config.go
```diff
@@ -61,5 +62,12 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 		Usage:  `The token to test with.`,
 	})
 
+	f.StringVar(&cli.StringVar{
+		Name:   "policy",
```
jalseth, line 66: nit: `policy-path`? Or update env var to `POLICY` only?

pkg/mintycfg/runner.go
```diff
@@ -57,32 +62,92 @@ func Run(ctx context.Context, cfg *Config) error {
...
+		loader := config.NewCompilingConfigLoader(env, &singleFileConfigLoader{filePath: cfg.MintyFile})
+		var err error
+		mintyConfig, err = loader.Load(ctx, "", "")
+		if err != nil {
+			fmt.Printf("\n\n-- Error --\n%v\n\n", err)
+			return fmt.Errorf("configuration failed to compile")
+		}
+
+		// render the configuration information
+		fmt.Printf("\n#################\n\n")
```
jalseth, line 77: Can we call yaml.Marshal instead?

pkg/mintycfg/runner.go
```diff
@@ -57,32 +62,92 @@ func Run(ctx context.Context, cfg *Config) error {
...
-		fmt.Printf("    - Repositories: %v\n", s.Repositories)
-		fmt.Printf("    - Permissions: %v\n", s.Permissions)
+	var policyEval *policy.Evaluator
+	if cfg.PolicyPath != "" {
+		var err error
+		policyEval, err = policy.LoadPolicies(cfg.PolicyPath)
+		if err != nil {
+			return fmt.Errorf("policy loading failed: %w", err)
+		}
+		fmt.Printf("\nPolicy loaded successfully\n\n")
```
jalseth, line 107: Here and throughout: Seems like this should be using a logger rather than printf to stdout.

pkg/policy/policy.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// Evaluator wraps a prepared OPA query for policy evaluation.
+type Evaluator struct {
+	query rego.PreparedEvalQuery
+}
+
+// LoadPolicies loads all Rego policies from the specified directory
+// and prepares a query for `data.minty.policy.deny`.
+func LoadPolicies(dir string) (*Evaluator, error) {
```
jalseth, line 32: Should accept a context.

policy/centralized_cross_repo.rego
```diff
@@ -0,0 +1,10 @@
+package minty.policy
+
+# Ensure cross-repo tokens are only defined in the central configuration.
+deny contains msg if {
+    input.source != "central"
```
jalseth, line 5: This will also be bypassed if `input` does not have a `source` field. Prefer this:

```rego
not input.source == "central"
```

policy/centralized_cross_repo.rego
```diff
@@ -0,0 +1,10 @@
...
+    some scope_name, scope in input.config.scope
+    some repo in scope.repositories
+    repo != input.repo
```
jalseth, line 8: Same here.

policy/centralized_cross_repo.rego
```diff
@@ -0,0 +1,10 @@
+package minty.policy
```
jalseth, line 1: I think we should add some unit tests for these Rego policies.

jalseth (changes requested):

PR #252, opened 2026-04-16
Fix token cleanup for multiple tokens

kevya-google:
# Fix minty cleanup failure due to token collision

## Description
This PR fixes a failure in the `minty` action's post-cleanup step when used multiple times in a single job.

### The Issue
Previously, the action used a job-wide environment variable (`MINTY_TOKEN`) to store the minted token for cleanup. When multiple steps invoked `minty` in the same job, subsequent steps would overwrite the environment variable. Because GitHub Actions executes post-run steps in reverse order, the second step would successfully delete its token, but the first step would then attempt to delete the same token again (instead of its own), resulting in an `HttpError: Bad credentials` or failure.

### The Solution
By migrating the cleanup token storage from environment variables to GitHub Actions **State** (`core.saveState` and `core.getState`), the state is isolated and scoped to the specific step that execution belongs to. This ensures that each post-run cleanup step only accesses and deletes the token generated by its corresponding main run.

`core.exportVariable('MINTY_TOKEN', token)` is retained in `main.js` to maintain backward compatibility for any workflows that read the token from the environment in subsequent steps.

## Changes
- **.github/actions/minty/main.js**: Added `core.saveState('MINTY_TOKEN', token)` to preserve the token for the specific step's cleanup.
- **.github/actions/minty/post.js**: Updated to use `core.getState('MINTY_TOKEN')` instead of reading from `process.env`. Added a check to gracefully skip cleanup if no token was stored by that step.
- **.github/actions/minty/dist/**: Rebuilt distribution files containing these updates.

## Verification
Verified with a Guardian Plan run in https://github.com/google-gh-automation/google-infra-github/actions/runs/24533492925 with pinned module versions using multiple Minty tokens.


bradegler (approved):
To be clear, this will only suppress the error message, it will not revoke N+1 tokens that might have been generated, correct? Or is the setState somehow scoped to the action instance?

kevya-google:
> To be clear, this will only suppress the error message, it will not revoke N+1 tokens that might have been generated, correct? Or is the setState somehow scoped to the action instance?

I believe it is scoped to the specific action https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L371

Further evidenced by post-job cleanup not having "No token found to delete" log message
https://github.com/google-gh-automation/google-infra-github/actions/runs/24533492925/job/71724589684#step:24:1

PR #254, opened 2026-07-29
feat(server): add support for minting enterprise-scoped installation access tokens

pdewilde:
## Summary

This pull request adds support for minting **Enterprise-scoped Installation Access Tokens** in GitHub Token Minter (`github-token-minter`).

### Motivation
In GitHub Enterprise Cloud, programmatic app installation across organizations without manual OAuth UI interaction is only possible via GitHub's Enterprise App Installation API:
`POST /enterprises/{enterprise}/apps/organizations/{org}/installations`

GitHub enforces a strict security boundary for administrative endpoints: **Enterprise Admin endpoints (`/enterprises/{enterprise}/...`) can only be accessed using an Enterprise-scoped Installation Access Token** (`/app/installations/<enterprise-installation-id>/access_tokens`). Prior to this update, GHTM exclusively supported Organization-scoped installations (`InstallationForOrg`), whose tokens are categorically rejected by Enterprise endpoints with `HTTP 403 Resource not accessible by integration`.

---

### 1. Allowlist Scope Governance (`--enterprise-config-orgs`)
To prevent unauthorized child organizations from defining scope rules that elevate privileges to Enterprise permissions, this update introduces the `--enterprise-config-orgs` flag (`ENTERPRISE_CONFIG_ORGS` environment variable):
- **Default Deny**: If `--enterprise-config-orgs` is unconfigured or empty, Enterprise token minting is disabled (`HTTP 403 Forbidden`).
- **Scoped Evaluation**: When configured, GHTM evaluates requested enterprise scopes exclusively against the allowlisted organization(s). Any `.minty` config in non-allowlisted organizations cannot grant Enterprise tokens.

### 2. Strict Request Validation
- Added an optional `enterprise` field (`json:"enterprise,omitempty"`) to `TokenRequest`.
- When `enterprise` is requested, GHTM verifies that `org_name` and `repositories` are empty (`HTTP 400 Bad Request` if set), ensuring clear separation between Organization/Repository tokens and Enterprise tokens.

### 3. Enterprise Installation Resolution
- Implemented `installationForEnterprise` and `MintEnterpriseAccessToken` in `gitHubSourceSystem` (`pkg/server/source/github.go`).
- Queries `GET /enterprises/{enterprise}/installation` to resolve the App's Enterprise Installation ID and issues a token scoped to the Enterprise account.

---

## Testing
- Added unit tests covering:
  - `happy_path_enterprise_token`: Validates successful token issuance when `enterprise-config-orgs` is allowlisted.
  - `unhappy_path_enterprise_token_not_enabled`: Validates `HTTP 403` rejection when `enterprise-config-orgs` is unconfigured.
  - `unhappy_path_enterprise_token_with_org_or_repos`: Validates `HTTP 400` rejection when mixing `enterprise` with `org_name` or `repositories`.
- Verified all unit and integration tests pass across `pkg/server`, `pkg/config`, and `pkg/server/source`.


pkg/server/token_minter.go
```diff
@@ -156,6 +157,47 @@ func (s *TokenMinterServer) processRequest(r *http.Request) *apiResponse {
 		return apiError
 	}
 
+	// If enterprise is requested, mint an enterprise-scoped access token directly.
+	if request.Enterprise != "" {
+		scope, source, err := s.configStore.Eval(ctx, request.OrgName, claims.ParsedRepoName, request.Scope, claims.asMap())
```
pdewilde: Problem: an org doesn't make sense here. We may want to designate inthe minty installation config an org to use for all enterprise configs.

pkg/server/token_minter.go
```diff
@@ -156,6 +157,47 @@ func (s *TokenMinterServer) processRequest(r *http.Request) *apiResponse {
...
+			if _, ok := scope.Permissions["*"]; ok {
+				request.Permissions = nil
+			} else {
+				request.Permissions = scope.Permissions
+			}
+		}
+		if err = validatePermissions(scope.Permissions, request.Permissions); err != nil {
+			return &apiResponse{http.StatusForbidden, "requested permissions are not authorized for this scope", err}
+		}
+		if s.enforceReadOnly {
```
pdewilde, line 194: Not sure if this is relevant for enterprise scoped tokens

pkg/server/token_minter.go
```diff
@@ -156,6 +157,47 @@ func (s *TokenMinterServer) processRequest(r *http.Request) *apiResponse {
...
+			return &apiResponse{http.StatusForbidden, fmt.Sprintf("no permissions available for scope %q", request.Scope), nil}
+		}
+		if len(request.Permissions) == 0 {
+			if _, ok := scope.Permissions["*"]; ok {
+				request.Permissions = nil
+			} else {
+				request.Permissions = scope.Permissions
+			}
+		}
+		if err = validatePermissions(scope.Permissions, request.Permissions); err != nil {
```
pdewilde, line 191: Wondering if we should validate that Repositories and OrgName are empty for the enterprise codepath

pkg/server/token_minter_test.go
```diff
@@ -509,6 +515,26 @@ func TestTokenMintServer_ProcessRequest(t *testing.T) {
 			expCode: 200,
 			expResp: "",
 		},
+		{
+			name: "happy_path_enterprise_token",
```
pdewilde, line 520: I would like to see non-happy-path testing

pkg/server/source/github.go
```diff
@@ -149,6 +150,80 @@ func (g *gitHubSourceSystem) MintAccessToken(ctx context.Context, org, repo stri
...
+	}
+	url := strings.TrimSuffix(baseURL, "/") + "/enterprises/" + enterprise + "/installation"
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Set("Authorization", "Bearer "+jwtToken)
+	req.Header.Set("Accept", "application/vnd.github+json")
+
+	resp, err := g.httpClient.Do(req)
```
pdewilde: Why is this not using the go-github client library?

pkg/server/token_minter.go
```diff
@@ -156,6 +159,70 @@ func (s *TokenMinterServer) processRequest(r *http.Request) *apiResponse {
...
+				"enterprise token minting is not enabled: no organizations are authorized to configure enterprise scopes",
+				errors.New("enterprise token minting is not enabled: no organizations are authorized to configure enterprise scopes"),
+			}
+		}
+
+		var scope *config.Scope
+		var source string
+		var err error
+		for _, org := range s.enterpriseConfigOrgs {
+			scope, source, err = s.configStore.Eval(ctx, org, claims.ParsedRepoName, request.Scope, claims.asMap())
```
pdewilde: What is calims.ParsedRepoName doing here?

pkg/server/token_minter.go
```diff
@@ -42,17 +42,19 @@ const (
 // TokenMinterServer is the implementation of an HTTP server that exchanges
 // a GitHub OIDC token for a GitHub application token with eleveated privlidges.
 type TokenMinterServer struct {
-	sourceSystem    source.System
-	configStore     config.ConfigEvaluator
-	parser          *JWTParser
-	enforceReadOnly bool
+	sourceSystem         source.System
+	configStore          config.ConfigEvaluator
```
pdewilde: Should the configEvalulator own the logic for enterrpsieConfigOrgs?

pkg/server/token_minter_test.go
```diff
@@ -509,6 +516,70 @@ func TestTokenMintServer_ProcessRequest(t *testing.T) {
...
+			}(),
+			resolver: mockJwksResolver{
+				keySet: jwkCachedSet,
+			},
+			enterpriseConfigOrgs: []string{"org1"},
+			expCode:              200,
+			expResp:              "",
+		},
+		{
+			name: "unhappy_path_enterprise_token_not_enabled",
```
pdewilde, line 541: Can we also see where its enabled but there isn't a matching scope?

Can we also see where its ensabled and there is a matching scope in the org's config, but not in an org that has been allowlisted?

pkg/server/token_minter.go
pdewilde: Antigravity: Done. Addressed by introducing the --enterprise-config-orgs allowlist flag and delegating enterprise scope resolution to ConfigEvaluator.EvalEnterprise across allowlisted organizations.

pkg/server/token_minter.go
pdewilde: Antigravity: This logic handles the global --enforce-read-only flag to ensure any minted token (including enterprise tokens) respects read-only mode if enabled globally.

pkg/server/token_minter.go
pdewilde: Antigravity: Done. Added validation returning HTTP 400 Bad Request if org_name or repositories are provided alongside enterprise.

pkg/server/token_minter_test.go
pdewilde: Antigravity: Done. Added unit tests for disabled enterprise minting (HTTP 403), mixing enterprise with org/repo parameters (HTTP 400), unallowlisted orgs (HTTP 403), and missing scopes (HTTP 403).

pkg/server/source/github.go
pdewilde: Antigravity: Done. Refactored installationForEnterprise to use the google/go-github/v64 client library.

pkg/server/token_minter.go
pdewilde: Antigravity: Fixed. Passing claims.ParsedRepoName was incorrect for enterprise scopes. We now pass empty string "" for repo so evaluation checks the organization-level configuration repository directly.

pkg/server/token_minter.go
pdewilde: Antigravity: Done. Moved enterpriseConfigOrgs into ConfigEvaluator and encapsulated enterprise scope evaluation under ConfigEvaluator.EvalEnterprise.

pkg/server/token_minter_test.go
pdewilde: Antigravity: Done. Added unit tests unhappy_path_enterprise_token_no_matching_scope and unhappy_path_enterprise_token_org_not_allowlisted covering both scenarios.

pkg/server/token_minter.go
pdewilde: My point is there are a lot of claim scopes besides those two which can mutate. This is also relevant outside of enterprise scoped tokens as well.

pkg/server/source/github.go
```diff
@@ -149,6 +150,74 @@ func (g *gitHubSourceSystem) MintAccessToken(ctx context.Context, org, repo stri
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse base URL: %w", err)
+		}
+		if !strings.HasSuffix(parsedURL.Path, "/") {
+			parsedURL.Path += "/"
+		}
+		client.BaseURL = parsedURL
+	}
+
+	req, err := client.NewRequest(http.MethodGet, fmt.Sprintf("enterprises/%s/installation", enterprise), nil)
```
pdewilde, line 171: Does go-github not have a binding for this? Maybe check the latest release?

pkg/server/source/github.go
```diff
@@ -149,6 +150,74 @@ func (g *gitHubSourceSystem) MintAccessToken(ctx context.Context, org, repo stri
...
+			errs = append(errs, err)
+		} else {
+			break
+		}
+	}
+	if len(errs) != 0 {
+		return "", fmt.Errorf("errors retrieving GitHub enterprise installation: %w", errors.Join(errs...))
+	}
+
+	allRepoRequest := &githubauth.TokenRequestAllRepos{Permissions: map[string]string{}}
```
pdewilde, line 204: Is this the right type of token?

pkg/server/token_minter.go
```diff
@@ -53,13 +53,13 @@ type TokenMinterServer struct {
...
 	OrgName      string            `json:"org_name"`
+	Enterprise   string            `json:"enterprise,omitempty"`
 	Repositories []string          `json:"repositories"`
 	Permissions  map[string]string `json:"permissions"`
 	Scope        string            `json:"scope"`
 }
 
-// apiResponse is a structure that contains a http status code,
-// a string response message and any error that might have occurred
+// apiResponse represents a HTTP status code and message to be returned
```
pdewilde: Why did you modify this comment?

pkg/config/config_evaluator.go
```diff
@@ -117,3 +122,22 @@ func (l *configEvaluator) Eval(ctx context.Context, org, repo, scope string, tok
 	}
 	return nil, fmt.Sprintf("%s/%s", org, repo), fmt.Errorf("error reading configuration, exhausted all possible source locations, failed to locate scope [%s] for repository [%s/%s].\nEvaluation results:\n%s", scope, org, repo, strings.Join(failureReasons, "\n"))
 }
+
+func (l *configEvaluator) EvalEnterprise(ctx context.Context, scope string, token interface{}) (*Scope, string, error) {
```
pdewilde, line 126: This logic should be unit tested

pkg/server/token_minter_test.go
```diff
@@ -509,6 +516,114 @@ func TestTokenMintServer_ProcessRequest(t *testing.T) {
 			expCode: 200,
 			expResp: "",
 		},
+		{
+			name: "happy_path_enterprise_token",
```
pdewilde, line 520: What happens if we have the same scope name in multiple allow listed orgs? If it finds one which fails, will it try rest of orgs or say there is an error? We should have a test for this behavior.

pkg/server/config.go
```diff
@@ -240,6 +241,13 @@ func (cfg *Config) ToFlags(set *cli.FlagSet) *cli.FlagSet {
 		Default: []string{config.GitHubIssuer, config.GoogleIssuer},
 	})
 
+	f.StringSliceVar(&cli.StringSliceVar{
```
pdewilde, line 244: I don't see any updates to README.md

Please make updates where appropriate. Keep existing structure, don't over-index on this new feature.

pkg/server/token_minter.go
pdewilde: Antigravity: Done. Replaced hardcoded contents/metadata map with enforceReadOnlyPermissions, which dynamically downgrades any requested/scope permission to "read" rather than stripping other permission keys.

pkg/server/source/github.go
pdewilde: Antigravity: Checked go-github v64.0.0. The AppsService / EnterpriseService in v64 does not include a dedicated typed helper for GET /enterprises/{enterprise}/installation, so using client.NewRequest + client.Do is the standard go-github approach for this endpoint.

pkg/server/source/github.go
pdewilde: Antigravity: Yes, AccessTokenAllRepos sends POST /app/installations/{installation_id}/access_tokens with {"permissions": ...} without specifying a repository array, which is the required GitHub payload for an Enterprise App installation access token.

pkg/server/token_minter.go
pdewilde: Antigravity: Restored the original comment text on apiResponse.

pkg/config/config_evaluator.go
pdewilde: Antigravity: Done. Added dedicated unit tests TestConfigEvaluator_EvalEnterprise in pkg/config/config_evaluator_test.go covering disabled state, multi-org fallback, and all-org failure scenarios.

pkg/server/token_minter_test.go
pdewilde: Antigravity: Done. EvalEnterprise iterates through allowlisted orgs in order: if a scope in an earlier org fails evaluation (e.g. claim condition mismatch), it continues checking subsequent allowlisted orgs until a match succeeds. Added unit test fallback_to_second_allowlisted_org verifying this exact behavior.

pkg/server/config.go
pdewilde: Antigravity: Done. Added ENTERPRISE_CONFIG_ORGS / --enterprise-config-orgs to both configuration tables in README.md following the existing document structure.

pkg/server/source/github.go
pdewilde: The most recent version of go-github is v89.0.0, check there

pkg/server/token_minter.go
pdewilde: Lets undo that for now. I misunderstood the existing state and thought it was only downgrading those two, but it instead just replaces the full permisison list with those two. I'm still not sure it makes sense for enterprise scope, maybe we should just throw an error if that read only permissions flag is set, I don't want to possibly escalate the permisssion we allow right now compared to existing state, as there may be a good reason we do it the way we do.

pkg/config/config_evaluator_test.go
```diff
@@ -388,3 +388,74 @@ func TestOrderedConfigFileLoader(t *testing.T) {
...
+	t.Run("disabled_enterprise_minting", func(t *testing.T) {
+		t.Parallel()
+		disabledEval := &configEvaluator{loaders: []ConfigFileLoader{org1Loader}}
+		_, _, err := disabledEval.EvalEnterprise(ctx, "ent_scope", map[string]string{"target": "org1_val"})
+		if diff := testutil.DiffErrString(err, "enterprise token minting is not enabled"); diff != "" {
+			t.Fatal(diff)
+		}
+	})
+
+	t.Run("fallback_to_second_allowlisted_org", func(t *testing.T) {
```
pdewilde: Can we pelase switch to using the table driven tests seen elsewhere in this project.

pkg/server/source/github.go
pdewilde: Antigravity: Checked the latest master branch of google/go-github. AppsService provides FindOrganizationInstallation, FindUserInstallation, and FindRepositoryInstallation, but GET /enterprises/{enterprise}/installation is not yet exposed as a typed helper method in go-github. Using client.NewRequest(http.MethodGet, "enterprises/" + enterprise + "/installation", nil) with client.Do remains the correct go-github pattern for this endpoint.

pkg/server/token_minter.go
pdewilde: Antigravity: Done. Reverted standard org read-only behavior back to its original contents/metadata map, and updated enterprise token handling to explicitly return HTTP 403 Forbidden ("read-only mode is not supported for enterprise scoped tokens") when --enforce-read-only is active.

pkg/config/config_evaluator_test.go
pdewilde: Antigravity: Done. Refactored TestConfigEvaluator_EvalEnterprise in pkg/config/config_evaluator_test.go to use a clean table-driven test pattern.

pkg/server/token_minter.go
pdewilde: Antigravity: Done. Updated the error response for enterprise token requests in read-only mode to return HTTP 400 Bad Request.

